PR #233, opened 2025-07-29
Draft: WIP

apryakhin:


src/realm/udp/udp_module.cc
```diff
@@ -0,0 +1,400 @@
...
+      NodeMeta meta;
+      meta.epoch = 1;
+      meta.ip = inet_addr(seed_ip.c_str());
+      meta.udp_port = seed_port;
+      Network::node_directory.add_slot(NodeDirectory::UNKNOWN_NODE_ID, meta);
+
+      register_peer(NodeDirectory::UNKNOWN_NODE_ID, seed_ip, seed_port);
+    }
+
+    rx_thread_ = std::thread([this]() { rx_loop(); });
```
eddy16112, line 121: It is better to use realm thread. You can take a look at the GPU worker https://gitlab.com/StanfordLegion/legion/-/blob/master/realm/src/realm/cuda/cuda_module.cc#L3449. The GPU worker can be either a dedicated thread or a bgwork item. If using realm thread, the runtime can control the binding along with other threads.

apryakhin: Done

src/realm/network.h
```diff
@@ -71,10 +76,29 @@ namespace Realm {
     // in most cases, there will be a single network module - if so, we set
     //  this so we don't have to do a per-node lookup
     extern NetworkModule *single_network;
+    extern NetworkModule* control_plane_network;
```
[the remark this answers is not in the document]
apryakhin: Right, that was the plan and I just haven't gotten that point yet

src/realm/node_directory.h
```diff
@@ -0,0 +1,105 @@
...
+#include "realm/event.h"
+#include "realm/nodeset.h"
+
+namespace Realm {
+  // -----------------------------------------------------------------
+  // Small immutable header describing a peer
+  // ------------------------------------------------------------------
+  struct NodeMeta {
+    NodeID id{NodeID(-1)};
+    uint64_t epoch{0};
```
eddy16112, line 20: Could you please add comments to explain what each field is?

src/realm/node_directory.h
```diff
@@ -0,0 +1,105 @@
...
+    uint32_t ip{0};
+    uint16_t udp_port{0};
```
eddy16112, line 22: Do you think if we can make the `NodeMeta` more generic? I mean not tied to UDP. At least not having a `dup_port`, but something like `ControlPlanPort control_port`, such that we can swap the UDP with something else in the future.

src/realm/runtime_impl.cc
```diff
@@ -1069,884 +1064,903 @@ namespace Realm {
     machine = new MachineImpl(this);
   }
 
-    RuntimeImpl::~RuntimeImpl(void)
-    {
-      delete machine;
-      delete core_reservations;
+  RuntimeImpl::~RuntimeImpl(void)
```
eddy16112, line 1067: Could you please not formatting the runtime_impl.cc? at least not formatting the lines that are not touched. Otherwise, there are so many changes in this file, and it is hard to see which one is the actual code change

src/realm/ucx/ucp_module.cc
```diff
@@ -142,7 +142,11 @@ namespace Realm {
     cp.add_option_string("-ucx:tls_dev", config.tls_dev);
 #endif
 
-    bool ok = cp.parse_command_line(cmdline);
+    cp.add_option_int_units("-ll:gsize", deprecated_gsize, 'm');
+    cp.add_option_int("-ll:id", config.rank_id);
```
eddy16112, line 146: Could you please explain what they are? why do we need to specify the rank_id?

src/realm/ucx/ucp_context.cc
```diff
@@ -452,17 +452,17 @@ namespace UCP {
     return false;
   }
 
-  bool UCPWorker::ep_get(int target, int remote_dev_index, ucp_ep_h *ep) const
```
eddy16112, line 455: Why removing the const? I do not think we modify the UCPWorker in this function?

src/realm/ucx/ucp_internal.cc
```diff
@@ -405,11 +407,24 @@ namespace UCP {
       status = UCP_FNPTR(ucp_worker_get_address)(wi.worker->get_ucp_worker(),
           &wi.addr, &wi.addrlen);
       CHKERR_JUMP(status != UCS_OK, "ucp_worker_get_address failed", log_ucp, err);
+
+      assert(config.rank_id == Network::my_node_id);
+      auto meta = *Network::node_directory.lookup(Network::my_node_id);
```
eddy16112, line 412: Is it possible to get rid of the global variable? It is hard to unit test.

src/realm/node_directory.h
```diff
@@ -0,0 +1,105 @@
...
+    std::atomic<uint8_t> state{0}; // 0 live, 1 retiring, 2 failed
+  };
+
+  // TODO:
+  // 1. It needs to be fast, concurent, scalable
+  // 2. It needs to integrate well with custom-built DHT
+  // and other external membership solutions such as etcd, zookeeper
+  // 3. Interface and implementation needs to be cleaned up
+
+  class NodeDirectory {
```
eddy16112, line 41: Could you please add some comments to the member function of NodeDirectory to explain what they are and when they need to be called.

src/realm/ucx/ucp_internal.cc
```diff
@@ -2381,8 +2423,16 @@ namespace UCP {
         goto err_update_pending;
       }
       *req = *req_prim;
-      CHKERR_JUMP(!worker->ep_get(target, remote_dev_index, &req->ucp.ep),
+
+      if (!worker->ep_get(target, remote_dev_index, &req->ucp.ep)) {
```
eddy16112, line 2427: Maybe we can move the check and add_remote_ep into `UCPMessageImpl::commit` instead of duplicating them in `commit_multicast` and `commit_unicast`.

src/realm/ucx/ucp_internal.h
```diff
@@ -104,8 +104,9 @@ namespace UCP {
     struct Config {
       AmWithRemoteAddrMode am_wra_mode{AM_WITH_REMOTE_ADDR_MODE_AUTO};
       bool bind_hostmem{true};
+      int rank_id{0};
       int pollers_max{2};
-      int num_priorities{2};
+      int num_priorities{1};
```
eddy16112, line 109: why do we need to change it?

src/realm/runtime_impl.cc
apryakhin: That's not the production change yet.

src/realm/ucx/ucp_module.cc
apryakhin: joining node needs to pick up a rank ID right? So, we specify it that will happen outside of realm scope.

src/realm/ucx/ucp_internal.cc
apryakhin: If you have a proposal on how to best approach it please propose

src/realm/ucx/ucp_internal.h
apryakhin: That's experimental - should be brought back

src/realm/ucx/ucp_internal.cc
```diff
@@ -880,6 +895,33 @@ namespace UCP {
...
+    const ucp_address_t*addr  = reinterpret_cast<const ucp_address_t*>(p);
+    const size_t        alen  = bytes - 1;     // address length (may be unused)
+
+    rdev = -1; // TODO: FIX ME
+
+    (void)alen;                                // silences “unused” in release
+
+    /* Choose host context, priority 0 – identical to create_eps() */
+    const UCPContext *ctx = get_context_host();
+    UCPWorker        *w   = get_tx_worker(ctx, /*priority=*/0);
```
eddy16112, line 913: Please do not forget to fix the `priority`, each context has multiple priorities, and each priority maps to a worker.

src/realm/ucx/ucp_internal.cc
eddy16112: Is it possible to pass the node_directory via the function arguments?

src/realm/membership/membership_mesh.cc
```diff
@@ -0,0 +1,385 @@
...
+
+    return realmStatus_t::REALM_SUCCESS;
+  }
+
+  // TODO: FIX ALL RACE CONDITIONS
+  realmStatus_t leave(void *st, const realmNodeMeta_t *self, realmMembershipHooks_t hooks)
+  {
+    MembershipMesh *state = static_cast<MembershipMesh *>(st);
+
+    state->hooks = hooks;
```
apryakhin, line 342: This is technically a bug - we should not re-assign the callbacks

src/realm/membership/membership_mesh.cc
```diff
@@ -0,0 +1,385 @@
...
+      am->announce_mm = msg.announce_mm;
+      am.add_payload(data, datalen);
+      am.commit();
+    }
+  }
+
+  // TODO: FIX ME - what if ACK is leaving as well? That will deadlock.
+  send_join_ack(sender, new_epoch, mesh_state->subscribers.size() + 1, msg.announce_mm);
+
+  if(mesh_state->hooks.filter) {
```
eddy16112, line 181: what does filter do?

src/realm/membership/membership_mesh.cc
```diff
@@ -0,0 +1,385 @@
...
+      hooks.pre_join(&meta, nullptr, 0, /*joined=*/false, hooks.user_arg);
+    }
+
+    if(self->seed_id == NodeDirectory::INVALID_NODE_ID) {
+      realmNodeMeta_t meta{(int32_t)Network::my_node_id, 0, announce_mm};
+      hooks.post_join(&meta, nullptr, 0, /*joined=*/true, hooks.user_arg);
+    } else {
+
+      ActiveMessage<JoinRequestMessage> am(self->seed_id, dbs.bytes_used());
+      am->epoch = Network::node_directory.cluster_epoch();
```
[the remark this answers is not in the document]
apryakhin: right but this value should not be hard coded, cluster_epoch() will return it.

src/realm/machine_impl.cc
```diff
@@ -3166,4 +3169,25 @@ namespace Realm {
     return chosen;
   }
 
+  void MachineImpl::remove_node(NodeID node_id)
```
apryakhin, line 3172: @eddy16112 can you look into this method please

src/realm/runtime_impl.cc
```diff
@@ -2318,6 +2353,166 @@ namespace Realm {
...
+
+      NodeMeta *nm = Network::node_directory.lookup(Network::my_node_id);
+      assert(nm != nullptr);
+
+      const uint8_t *buffer = static_cast<const uint8_t *>(dbs.get_buffer());
+      nm->machine_model.assign(buffer, buffer + dbs.bytes_used());
+
+      realmNodeMeta_t self_meta{};
+      self_meta.node_id = Network::my_node_id;
+      self_meta.seed_id = Network::my_node_id == 0 ? NodeDirectory::INVALID_NODE_ID
```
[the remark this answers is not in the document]
apryakhin: Yes, it is. This is the zero rank for which the seed isn't provided. It has to be detected programmatically and done properly. That's just a hack for now

PR #237, opened 2025-08-06
Draft: WIP

apryakhin:


realm/src/realm/udp/udp_shim.h
```diff
@@ -0,0 +1,80 @@
...
+  // for UDP transport.  This code is transport-agnostic: the owner supplies a
+  // callback that actually transmits the byte-vector.
+  //
+  // Packet format expected by the caller:
+  //   | UDPHeader | user_header | payload |
+  // where UDPHeader contains `seq` and `ack_bits` fields (see udp_module.h).
+  // UDPShim merely fills those two fields and manages retransmission.
+  //
+  // Lifetime: one UDPShim instance per peer.
+  class UDPShim {
```
apryakhin, line 22: @eddy16112

realm/src/realm/ucx/ucp_internal.cc
```diff
@@ -1672,6 +1742,8 @@ namespace Realm {
           total_rcomp_sent.load(), total_rcomp_received.load(), outstanding_reqs.load(),
       };
 
+      return true; // TODO: FIX ME
```
apryakhin, line 1745: quiescence is broken since that requires all_gather and we aren't maintaining ucc communicators (yets) with UDP control plane.

realm/src/realm/node_directory.h
```diff
@@ -0,0 +1,101 @@
...
+    public:
+      virtual ~Provider() = default;
+      virtual void put(NodeSet peers, const void *blob, size_t bytes,
+                       uint64_t ttl_sec = 0) = 0;
+      virtual void fetch(NodeID id) = 0;
+    };
+
+    void set_provider(Provider *_provider) { provider = _provider; }
+
+    uint64_t bump_epoch(NodeID id);
```
apryakhin, line 68: This is just for POC and that needs to be properly done to guarantee unique epoch here

realm/src/realm/runtime_impl.cc
```diff
@@ -1913,18 +2007,18 @@ namespace Realm {
     activemsg_handler_table.construct_handler_table();
 
     // and also our incoming active message manager
-    message_manager = new IncomingMessageManager(
-        Network::max_node_id + 1, config->active_msg_handler_threads, *core_reservations);
+    message_manager = new IncomingMessageManager(256, config->active_msg_handler_threads,
```
apryakhin, line 2010: TODO: we need to ensure that incoming message manager handles properly both networks or/and we need one per network.

realm/src/realm/quiescence.cc
```diff
@@ -0,0 +1,241 @@
...
+    const uint64_t epoch = s_node_dir->cluster_epoch();
+
+    assert(!members.contains(leaving));
+
+    {
+      AutoLock<> al(quiesce_state.mtx);
+      if(quiesce_state.round_id == 0) {
+
+        QuiescenceCounters local_counters;
+        collect_all_counters(leaving, local_counters);
```
apryakhin, line 194: here is the bug we need to collect counters in a loop for every peer.

PR #240, opened 2025-08-06
cuda: Fixes for 13.0

muraj:
* cuGetProcAddress broke it's own rules for a couple of apis and kept back old versions of apis, but left cuGetProcAddress to return the newer ones, breaking source compatibility guarentees for 13.0
* For those apis that were held back, an explicit context or location was added, so modify the calls to pass in the required arguments, being compatible with older toolkits as well.

### DIFF
--- a/src/realm/cuda/cuda_internal.h
+++ b/src/realm/cuda/cuda_internal.h
@@ -1278,6 +1278,20 @@ namespace Realm {
     CUresult cuCtxRecordEvent(CUcontext hctx, CUevent event);
 #endif
 
+#if CUDA_VERSION >= 13000
+// Unfortunately, 13.0 violates it's own source compatibility rules versus
+// cuGetProcAddress, so fix that ourselves here.
+#if !defined(cuCtxGetDevice)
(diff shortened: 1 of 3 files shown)


src/realm/cuda/cuda_internal.h
```diff
@@ -1278,6 +1278,20 @@ namespace Realm {
     CUresult cuCtxRecordEvent(CUcontext hctx, CUevent event);
 #endif
 
+#if CUDA_VERSION >= 13000
+// Unfortunately, 13.0 violates it's own source compatibility rules versus
+// cuGetProcAddress, so fix that ourselves here.
+#if !defined(cuCtxGetDevice)
```
eddy16112, line 1284: If there is no `define cuCtxGetDevice cuCtxGetDevice_v2` in cuda 13, why do we need `!defined(cuCtxGetDevice)`?

eddy16112 (approved):

src/realm/cuda/cuda_internal.h
muraj: In case they add it, like in 13.1 or something.

PR #245, opened 2025-08-11
Add event_timedwait to the realm CAPI.

ebendler:


src/realm/realm_c.h
```diff
@@ -776,6 +793,18 @@ realm_status_t REALM_EXPORT realm_user_event_trigger(realm_runtime_t runtime,
...
+/**
+ * @brief Cancels a user event.
+ *
+ * @param runtime The runtime instance to use.
+ * @param event The user event to be cancelled.
+ * @return Realm status indicating success or failure.
+ *
+ * @ingroup UserEvent
+ */
+realm_status_t REALM_EXPORT realm_user_event_cancel(realm_runtime_t runtime,
```
muraj, line 805: I would like this not to be separate from realm_event_cancel_operation.  I know the C++ api separates these things, but in the C API I would like these to be one in the same and we'd just figure it out internally which path to go down.  This will probably require some larger refactoring internally to support (for example, modifying Event::cancel_operation to, if no operation is found, trigger itself with a poisoning, which will need to be tested).

src/realm/realm_c.h
```diff
@@ -745,6 +745,22 @@ realm_status_t REALM_EXPORT realm_event_has_triggered(realm_runtime_t runtime,
...
+ * @param runtime The runtime instance to use.
+ * @param event The event to wait for.
+ * @param max_ns The maximum number of nanoseconds to wait.
+ * @param[out] has_triggered Whether the event has triggered.
+ * @param[out] poisoned Whether the event is poisoned.
+ * @return Realm status indicating success or failure.
+ *
+ * @ingroup Event
+ */
+realm_status_t REALM_EXPORT realm_event_timedwait(realm_runtime_t runtime,
```
muraj, line 760: Can we instead add the time to realm_event_wait instead of having a separate call?  Then we can just pass zero or something for an infinite wait, which is what realm_event_wait does already.  I'm usually not in favor of modifying already established C APIs, and we'll need to be more strict about this eventually, but since the C API is still experimental, I think we're okay.

muraj (changes requested):

src/realm/realm_c.h
ebendler: ok

src/realm/realm_c.h
ebendler: ok, I'll leave that for another PR then

muraj: @eddy16112 what difference does that make?  In the end, both events and user events are triggered in the poisoned state.  I don't see why we have to make this distinction.  If I'm given an event, regardless of it being a user event or a generic event, I should be able to cancel it.  Right now, with this API split, my application needs to carry the information of whether it is a user event or a normal event to call the right function.  But these handles are inter-changable, and even in the C++ api, you can call cancel_operation on a UserEvent, and it actually doesn't do anything, which IMHO is a bug.  We should fix this and unify both paths.

muraj: @ebendler just historical I believe.

ebendler: I put in realm_event_cancel_operation as an API that will just trigger an event with the poisoned status. This is how the UserEvent cancel worked in the old API, while Event.Cancel_Operation is doing a bunch of other stuff.

src/realm/realm_c.h
```diff
@@ -705,13 +705,17 @@ realm_status_t REALM_EXPORT realm_memory_query_iter(realm_memory_query_t query,
...
+ * @param max_ns The maximum number of nanoseconds to wait. Any value <= 0 means wait
+ * forever.
+ * @param[out] has_triggered Whether the event has triggered. Only valid if max_ns > 0.
  * @param[out] poisoned Whether the event is poisoned.
  * @return Realm status indicating success or failure.
  *
  * @ingroup Event
  */
 realm_status_t REALM_EXPORT realm_event_wait(realm_runtime_t runtime, realm_event_t event,
+                                             long long max_ns, int *has_triggered,
```
muraj, line 718: Let's not add "has_triggered".  An application can always query directly realm_event_has_triggered.

src/realm/realm_c.cc
```diff
@@ -807,6 +786,74 @@ realm_status_t realm_event_has_triggered(realm_runtime_t runtime, realm_event_t
...
+    return status;
+  }
+
+  status = check_event_validity(event);
+  if(status != REALM_SUCCESS) {
+    return status;
+  }
+
+  // has_triggered is required only when doing a timed wait (max_ns > 0)
+  if(max_ns > 0 && has_triggered == nullptr) {
```
muraj, line 804: why max_ns > 0?  Any value for max_ns is valid.

src/realm/realm_c.cc
```diff
@@ -807,6 +786,74 @@ realm_status_t realm_event_has_triggered(realm_runtime_t runtime, realm_event_t
...
+  Realm::RuntimeImpl *runtime_impl = nullptr;
+  realm_status_t status = check_runtime_validity_and_assign(runtime, runtime_impl);
+  if(status != REALM_SUCCESS) {
+    return status;
+  }
+  status = check_event_validity(event);
+  if(status != REALM_SUCCESS) {
+    return status;
+  }
+  Realm::GenEventImpl::trigger(Realm::Event(event), true /*poisoned*/);
```
muraj, line 853: Hmmm... no... this is incorrect.  Triggering an event like this will not cancel the operation that the event is supposed to be triggered by (i.e. from a spawn() call).  We need to call Realm::Event::cancel_operation.  But Realm::Event::cancel_operation needs to, if there is no operation associated with the event that will trigger the poisoning for it, trigger itself with poisoning.  This in effect deals with userevents as well as events created through merges.

src/realm/realm_c.cc
muraj: Where are the tests in c_event.cc for these specific cases?  I.E. where max_ns > 0, etc?

src/realm/realm_c.h
```diff
@@ -746,6 +750,22 @@ realm_status_t REALM_EXPORT realm_event_has_triggered(realm_runtime_t runtime,
                                                       realm_event_t event,
                                                       int *has_triggered, int *poisoned);
 
+/**
+ * @brief Cancels the operation associated with an event.
```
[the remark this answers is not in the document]
ebendler: done

muraj (changes requested):

src/realm/realm_c.h
eddy16112: Apparently, all event will have an associated operation, if not, it could be a remote event https://gitlab.com/StanfordLegion/legion/-/blob/master/realm/src/realm/event_impl.cc?ref_type=heads#L349-364. I think that's why the cancel_operation and cancel are different.

src/realm/realm_c.h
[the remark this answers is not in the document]
ebendler: So...cancel_operation actually _doesn't_ poison the event as currently implemented. Is this a behavior change we want to make?

src/realm/event_impl.cc
```diff
@@ -359,8 +359,10 @@ namespace Realm {
         op->remove_reference();
       }
     } else {
-      // Slow path, triggering operation is remote
-      get_runtime()->optable.request_cancellation(*this, reason_data, reason_len);
```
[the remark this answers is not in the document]
ebendler: updated.

PR #248, opened 2025-08-12
Various CI fixes

muraj:
* Add sanitizer tests for ASAN, TSAN, and UBSAN
  * Unfortunately we have a leak in the python module, so disable that.
  * Fix a couple of found leaks
* Add ucx and gasnet builds to the ci


.github/workflows/legion-patches/0001-legion-fix-trie-test-to-properly-include-cstdint.patch
eddy16112, line 1: If it is a patch to legion, can we make a PR to update the Legion repo directly?

eddy16112 (approved):

PR #256, opened 2025-08-20
Refactor realm_event_wait to take a timer parameter

ebendler:
This eliminates the need for timed_wait as a separate function. Then update the tests to the new API definition

### DIFF
--- a/src/realm/realm_c.cc
+++ b/src/realm/realm_c.cc
@@ -824,23 +824,40 @@ realm_status_t realm_memory_query_iter(realm_memory_query_t query,
 /* Event API */
 
 realm_status_t realm_event_wait(realm_runtime_t runtime, realm_event_t event,
-                                int *poisoned)
+                                int64_t max_ns, int *poisoned)
 {
   Realm::RuntimeImpl *runtime_impl = nullptr;
   realm_status_t status = check_runtime_validity_and_assign(runtime, runtime_impl);
   if(status != REALM_SUCCESS) {
     return status;
   }
+
   status = check_event_validity(event);
   if(status != REALM_SUCCESS) {
     return status;
   }
+
+  // special case: NO_EVENT is always triggered
+  if(event == REALM_NO_EVENT) {
+    if(poisoned != nullptr) {
+      *poisoned = 0;
+    }
+    return REALM_SUCCESS;
+  }
+
   Realm::Event cxx_event = Realm::Event(event);
   bool poisoned_cxx = false;
-  cxx_event.wait_faultaware(poisoned_cxx);
+
+  if(max_ns == REALM_WAIT_INFINITE) {
+    cxx_event.wait_faultaware(poisoned_cxx);
+  } else {
+    cxx_event.external_timedwait_faultaware(poisoned_cxx, max_ns);
+  }
+
   if(poisoned != nullptr) {
     *poisoned = poisoned_cxx ? 1 : 0;
   }
+
   return REALM_SUCCESS;
 }
 
--- a/src/realm/realm_c.h
+++ b/src/realm/realm_c.h
@@ -173,6 +173,8 @@ typedef struct realm_region_instance_copy_params_t {
 #define REALM_TASK_ID_PROCESSOR_SHUTDOWN ((realm_task_func_id_t)2U)
 #define REALM_TASK_ID_FIRST_AVAILABLE ((realm_task_func_id_t)4U)
 
+#define REALM_WAIT_INFINITE ((int64_t)INT64_MIN)
+
 typedef enum realm_register_task_flags
 {
   REALM_REGISTER_TASK_DEFAULT = 0x0ULL,
@@ -758,14 +760,16 @@ realm_status_t REALM_EXPORT realm_memory_query_iter(realm_memory_query_t query,
  * @brief Waits for a specific event to complete.
  *
  * @param runtime The runtime instance to use.
- * @param[out] event The event to wait for.
+ * @param event The event to wait for.
+ * @param max_ns The maximum number of nanoseconds to wait.
+ *               REALM_WAIT_INFINITE is a special value that means wait forever.
  * @param[out] poisoned Whether the event is poisoned.
  * @return Realm status indicating success or failure.
  *
  * @ingroup Event
  */
 realm_status_t REALM_EXPORT realm_event_wait(realm_runtime_t runtime, realm_event_t event,
-                                             int *poisoned);
+                                             int64_t max_ns, int *poisoned);
 
 /**
  * @brief Merges multiple events into a single event.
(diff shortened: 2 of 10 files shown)


src/realm/realm_c.cc
```diff
@@ -733,23 +733,54 @@ realm_status_t realm_memory_query_iter(realm_memory_query_t query,
 /* Event API */
 
 realm_status_t realm_event_wait(realm_runtime_t runtime, realm_event_t event,
-                                int *poisoned)
+                                long long max_ns, int *has_triggered, int *poisoned)
```
muraj, line 736: Don't need has_triggered, this can be queried afterward with a separate api call.

src/realm/realm_c.cc
```diff
@@ -733,23 +733,54 @@ realm_status_t realm_memory_query_iter(realm_memory_query_t query,
...
     return status;
   }
+
   status = check_event_validity(event);
   if(status != REALM_SUCCESS) {
     return status;
   }
+
+  // has_triggered is required only when doing a timed wait (max_ns > 0)
+  if(max_ns > 0 && has_triggered == nullptr) {
```
muraj, line 750: Even if we decide to keep has_triggered, we shouldn't predicate it's usefulness on max_ns.

src/realm/realm_c.cc
muraj: max_ns should also be unsigned.

src/realm/realm_c.cc
```diff
@@ -733,23 +733,54 @@ realm_status_t realm_memory_query_iter(realm_memory_query_t query,
...
+    }
+    return REALM_SUCCESS;
+  }
+
   Realm::Event cxx_event = Realm::Event(event);
   bool poisoned_cxx = false;
-  cxx_event.wait_faultaware(poisoned_cxx);
+  bool triggered = false;
+
+  if(max_ns <= 0) {
```
muraj, line 769: max_ns == 0 is valid for peeking events, so zero should still go down the timedwait path.

Specifying a define for infinite wait would be a better way to approach this, like ULONGLONG_MAX or something.

tests/c/test_event.cc
muraj, line 1: Where are the tests for these new parameters?  All I see here are for zero and nullptr...

muraj (changes requested):

tests/c/test_event.cc
ebendler: They're in a new file that didn't get added

eddy16112 (commented):
Just curious why do you want to add the timedwait now? Does FlashCache use it? If we want to add it, it is better to test it in both integration and unit test. We could not test the wait in unit test now, but at least we can add some DISABLED tests, and enable them once we can make the event wait working.

src/realm/realm_c.cc
ebendler: ok

ebendler:
> Just curious why do you want to add the timedwait now? Does FlashCache use it? If we want to add it, it is better to test it in both integration and unit test. We could not test the wait in unit test now, but at least we can add some DISABLED tests, and enable them once we can make the event wait working.

I believe that I needed it for one of the later tests in the other branch.

tests/unit_tests/c_event.cc
```diff
@@ -195,10 +195,25 @@ TEST_F(CEventTest, DISABLED_EventWaitNotTriggeredEvent)
   realm_runtime_t runtime = *runtime_impl;
   ASSERT_REALM(realm_user_event_create(runtime, &event));
 
-  realm_status_t status = realm_event_wait(runtime, event, nullptr);
+  realm_status_t status = realm_event_wait(runtime, event, INT64_MAX, nullptr);
   EXPECT_EQ(status, REALM_SUCCESS);
 }
 
+TEST_F(CEventTest, DISABLED_EventWaitTimeoutEvent)
```
eddy16112, line 202: We might need to have a test case that triggering the event within the max_ns, however, TBH, I have no idea how to implement it within the gtest. Since the wait will block the calling thread, we will need to call the trigger in another thread

eddy16112 (commented):
LGTM, except for the comment I left for the unit test.

src/realm/realm_c.cc
```diff
@@ -824,23 +824,40 @@ realm_status_t realm_memory_query_iter(realm_memory_query_t query,
...
     return status;
   }
+
   status = check_event_validity(event);
   if(status != REALM_SUCCESS) {
     return status;
   }
+
+  // special case: NO_EVENT is always triggered
+  if(event == REALM_NO_EVENT) {
```
muraj, line 841: I believe you're going to want to move this up before check_event_validity(event), as this will return invalid event for NO_EVENT IIRC.  But yes, _this_ particular call should return SUCCESS.

src/realm/realm_c.cc
```diff
@@ -824,23 +824,40 @@ realm_status_t realm_memory_query_iter(realm_memory_query_t query,
...
+      *poisoned = 0;
+    }
+    return REALM_SUCCESS;
+  }
+
   Realm::Event cxx_event = Realm::Event(event);
   bool poisoned_cxx = false;
-  cxx_event.wait_faultaware(poisoned_cxx);
+
+  if(max_ns == INT64_MAX) {
```
muraj, line 851: Let's make this a define somewhere in realm_c.h so it's clear.  Something like `#define REALM_WAIT_INFINITE (INT64_MIN)` or something if you want.

tests/unit_tests/c_event.cc
```diff
@@ -195,10 +195,25 @@ TEST_F(CEventTest, DISABLED_EventWaitNotTriggeredEvent)
...
+{
+  realm_user_event_t event = REALM_NO_EVENT;
+  realm_runtime_t runtime = *runtime_impl;
+  ASSERT_REALM(realm_user_event_create(runtime, &event));
+
+  realm_status_t status = realm_event_wait(runtime, event, 10000, nullptr);
```
muraj, line 208: Probably need to use std::chrono::clock to take timestamps before and after and just ensure that the difference in timestamps is at least as large as the requested timeout.

muraj (changes requested):
Change looks good, couple of minor fixes and I can approve.

src/realm/realm_c.cc
muraj: Ah, missed that, you're right.

src/realm/realm_c.h
```diff
@@ -173,6 +173,8 @@ typedef struct realm_region_instance_copy_params_t {
 #define REALM_TASK_ID_PROCESSOR_SHUTDOWN ((realm_task_func_id_t)2U)
 #define REALM_TASK_ID_FIRST_AVAILABLE ((realm_task_func_id_t)4U)
 
+#define REALM_WAIT_INFINITE INT64_MIN
```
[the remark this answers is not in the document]
ebendler: done

muraj (approved):

PR #260, opened 2025-08-21
Support transfers for layouts with multiple id-indexed fields (PR 1/6)

2dm:
This is the first PR out of 6 and establishes the basic building blocks.

This patch adds a fast‐path for multi‐field copies when source and destination instances store fields consecutively in memory with strictly increasing IDs. Instead of the generic per‐field iterator, Realm switches to an IDIndexedIterator and issues the entire copy as a single batched DMA operation.
Fast‐Path Mechanics

Layout Detection: During instance construction, Realm identifies when field IDs form a consecutive sequence and marks the instances for the optimized path.

IDIndexedIterator: Attaches a single FieldBlock (allocated from the replicated heap) to the address list, skipping per‐field offset lookups.

Affine Collapse: Treats the multi‐field operation as one affine rectangle with N attached fields, collapsing what would be N separate copies into a single batched command.

Flow‐Control Splitting: The DMA engine automatically splits the batch according to internal flow‐control limits and IB fragment sizes (for cross‐node transfers).

Testing

Integration Test: Verifies end-to-end multi-field copy behavior on real instances.

Unit Tests: Covers iterator construction, address-list setup, FieldBlock allocation, and DMA splitting logic.

```
src_inst{ 0: [..field_size..], 1: [..field_size..], 2: [..field_size..], N-1: [..field_size..] }
dst_inst{ 0: [..field_size..], 1: [..field_size..], 2: [..field_size..], N-1: [..field_size..] }
index_space.copy(src_fids[32, 16, 612, 0..], dst_fids[17, 3, 999, 2..])
```


src/realm/cuda/cuda_memcpy.cu
```diff
@@ -132,16 +133,77 @@ static __device__ inline void memcpy_kernel_transpose(
 
 #define MAX_UNROLL (1)
 
-template <typename T, size_t N, typename Offset_t = size_t>
+/*template <typename T, size_t N, typename Offset_t = size_t>
```
apryakhin, line 136: This must some pointless change with commenting out the code..should be removed

src/realm/cuda/cuda_internal.cc
```diff
@@ -311,13 +320,137 @@ namespace Realm {
...
+        copy_info.extents[2] = planes;
+        copy_info.volume = planes * lines * contig_bytes;
+      }
+
+      in_alc.advance(id, planes * iscale);
+      out_alc.advance(od, planes * oscale);
+      return planes * lines * contig_bytes;
+    }
+
+    /*static size_t read_address_entry(AffineCopyInfo<3> &copy_infos, size_t &min_align,
```
apryakhin, line 449: should be removed

PR #269, opened 2025-08-25
add C API for profiling

eddy16112:


src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
...
+      delete m;
+      return status;
+    } else {
+      return REALM_PROFILING_ERROR_INVALID_MEASUREMENT;
+    }
+  }
+}
+
+realm_status_t
+realm_profiling_response_get_measurement(const realm_profiling_response_t *response,
```
muraj, line 1836: This is looking a lot better now.

src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
...
+    return REALM_PROFILING_ERROR_INVALID_MEASUREMENT;
+  }
+
+  const Realm::ProfilingMeasurements::OperationCopyInfo::InstInfo &inst_info =
+      m.inst_info[index];
+  const std::vector<Realm::FieldID> &field_vec =
+      (measurement_id >= PMID_OP_COPY_INFO_DST_FIELD ? inst_info.dst_fields
+                                                     : inst_info.src_fields);
+
+  *result_count = field_vec.size();
```
muraj, line 1807: You need to check that the result_count >= field_vec.size() here.  This needs to look like this from the application perspective:

```c++
size_t count = 0;
std::vector<realm_prof_inst_info_t> inst_info;
get_measurement(response, ID, nullptr, &count);
inst_info.resize(count);
// Notice that count is passed in again!
get_measurement(response, ID, inst_info.data(), &count);
```

This is to ensure we don't buffer overrun the provided buffer.

src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
...
+  if(result != nullptr) {
+    memcpy(result, field_vec.data(), field_vec.size() * sizeof(Realm::FieldID));
```
muraj, line 1809: Don't use `sizeof(Realm::FieldID)`, use `sizeof(field_vec[0])` to make sure the size follows the type you're actually copying in case it changes in the future.

src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
...
+    const Realm::ProfilingMeasurements::OperationCopyInfo &m, realm_field_id_t *result,
+    size_t *result_count, const void *data)
+{
+  realm_profiling_measurement_id_t measurement_id =
+      *(static_cast<const realm_profiling_measurement_id_t *>(data));
+  int index = measurement_id - (measurement_id >= PMID_OP_COPY_INFO_DST_FIELD
+                                    ? PMID_OP_COPY_INFO_DST_FIELD
+                                    : PMID_OP_COPY_INFO_SRC_FIELD);
+
+  if(index < 0 || static_cast<size_t>(index) >= m.inst_info.size()) {
```
muraj, line 1797: () around binary operators in if statement:

```c++
if ((index < 0) || (static_cast<size_t>(index) >= m.inst_info.size()) { }
```

src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
   *memory = Realm::Memory(external_resource_cxx->suggested_memory());
   return REALM_SUCCESS;
 }
+
+// Profiling API
+
+static inline realm_status_t
+copy_measurement(const Realm::ProfilingMeasurements::OperationBacktrace &m,
```
muraj, line 1604: You know, thinking about it, it might be a good idea to move these into the classes themselves.  Like have a OperationBacktrace::to_c(void *data, size_t n).  Then this can be moved to profiling.h.

When it comes time to implement this in realm.hpp, we won't need all this, since we can just keep the c structures inside the c++ class.

muraj (changes requested):

src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
...
+static inline realm_status_t
+handle_measurement(const Realm::ProfilingResponse &response_cxx,
+                   C_Measurement_Type *result, size_t *result_count,
+                   const void *data = nullptr)
+{
+  if(result_count == nullptr) {
+    bool has = response_cxx.has_measurement<CXX_Measurement_Type>();
+    return has ? REALM_SUCCESS : REALM_PROFILING_ERROR_INVALID_MEASUREMENT;
+  } else {
+    CXX_Measurement_Type *m = response_cxx.get_measurement<CXX_Measurement_Type>();
```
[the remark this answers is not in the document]
eddy16112: good point.

src/realm/realm_c.cc
eddy16112: fixed.

src/realm/realm_c.cc
eddy16112: fixed

src/realm/realm_c.cc
eddy16112: fixed

src/realm/realm_c.cc
eddy16112: This is actually a great idea. I will work on it.

src/realm/realm_c.cc
```diff
@@ -1535,3 +1597,347 @@ realm_status_t realm_external_resource_suggested_memory(
...
+    status =
+        handle_measurement<Realm::ProfilingMeasurements::OperationCopyInfo,
+                           realm_profiling_measurement_operation_copy_info_inst_info_t>(
+            response_cxx,
+            static_cast<realm_profiling_measurement_operation_copy_info_inst_info_t *>(
+                result),
+            result_count);
+    break;
+  default:
+    if(measurement_id >= PMID_OP_COPY_INFO_SRC_INST &&
```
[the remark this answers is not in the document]
eddy16112: done.

src/realm/realm_c.h
```diff
@@ -280,6 +277,145 @@ typedef enum realm_file_mode_t
...
+
+// The src and dst instances and fields are not included in the inst_info, because they
+// are retrieved from the PMID_OP_COPY_INFO_SRC_INST, PMID_OP_COPY_INFO_DST_INST,
+// PMID_OP_COPY_INFO_SRC_FIELD, PMID_OP_COPY_INFO_DST_FIELD.
+typedef struct realm_profiling_measurement_operation_copy_info_inst_info_st {
+  realm_region_instance_t src_indirection_inst; // src indirection instance (gather)
+  realm_region_instance_t dst_indirection_inst; // dst indirection instance (scatter)
+  realm_field_id_t src_indirection_field;       // field of indirection points
+  realm_field_id_t dst_indirection_field;       // field of indirection points
+  realm_profiling_measurement_operation_copy_info_request_type_t
```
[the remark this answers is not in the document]
eddy16112: Yeah, the name is too long. I renamed them to realm_profiling_x_y_z, just because personally I would prefer the full English words as it is easy to understand.

src/realm/realm_c.h
```diff
@@ -280,6 +277,145 @@ typedef enum realm_file_mode_t
   LEGION_FILE_CREATE = REALM_FILE_CREATE,
 } realm_file_mode_t;
 
+// Profiling measurement IDs
+typedef enum realm_profiling_measurement_id_enum
```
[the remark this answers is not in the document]
eddy16112: done

src/realm/realm_c.h
```diff
@@ -280,6 +277,145 @@ typedef enum realm_file_mode_t
...
+  size_t data_size;
+} realm_profiling_response_t;
+
+// When profiling the backtrace, the maximum length of a symbol supported by Realm is 2048
+// bytes
+#define PMID_OP_BACKTRACE_SYMBOLS_MAX_LENGTH 2048
+
+typedef struct realm_profiling_measurement_operation_backtrace_symbol_st {
+  char symbol[PMID_OP_BACKTRACE_SYMBOLS_MAX_LENGTH];
+} realm_profiling_measurement_operation_backtrace_symbol_t;
```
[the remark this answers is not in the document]
eddy16112: because in c++, it is std::vector<string>. Here, the caller will pass a `realm_profiling_measurement_operation_backtrace_symbol_t*`, otherwise, they will have to pass an array of `char symbol[PMID_OP_BACKTRACE_SYMBOLS_MAX_LENGTH];`

src/realm/realm_c.h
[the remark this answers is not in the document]
muraj: Documentation makes it easier to understand.  Most codes will shorten the type names significantly.  We don't need full english names.  Like in this case, the following is enough:

```c
realm_prof_copy_info_t
```

There's no need to add `request_type` or `measurement_operation`, as it is just redundant and doesn't add any value that documentation wouldn't give.

src/realm/realm_c.h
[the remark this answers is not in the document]
eddy16112: But how should the caller know they need to use `char symbol[PMID_OP_BACKTRACE_SYMBOLS_MAX_LENGTH];`. I think there are many advantages when warp it into a struct. 
1. Type safety. if we define a `realm_profiling_measurement_operation_backtrace_symbol_t symbols[N];`, then symbols[i] is clearly a symbol struct, not just a char*. Otherwise, the users will have to create an array of char[N * PMID_OP_BACKTRACE_SYMBOLS_MAX_LENGTH], and they have to carefully jump `PMID_OP_BACKTRACE_SYMBOLS_MAX_LENGTH` to access each symbol. 
2. Encapsulation and readability. By wrapping the array in a struct, you can give it a meaningful name (symbol), making the code easier to read and maintain. Using an array of array is really hard to read. 
3. Compatibility with STL containers. Structs can be stored in std::vector, std::array, or other containers easily, e.g. `std::vector< realm_profiling_measurement_operation_backtrace_symbol_t > symbols;`

I do not understand why you want to remove this struct.

src/realm/realm_c.h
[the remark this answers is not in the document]
eddy16112: The `measurement` isn’t very useful, so I removed it. However, it’s better to keep `operation` (or `op`)—I prefer `operation` because we generally use full names elsewhere, e.g., `instance`, `processor` and etc, and this keeps the naming consistent. Having `operation` makes it immediately clear that the profiling data comes from an operation. Similarly, profiling entries that start with `instance` indicate that the data comes from instance creation or redistribution. Honestly, I think it’s easier to understand how to use the profiling measurements this way, rather than constantly referring back to the documentation to remember what each class represents. That’s why the C++ naming also starts with `Operation` or `Instance`.

Regarding `realm_profiling_operation_copy_info_request_type_t`, I’m not sure how to shorten the name while still clearly indicating that it relates to `copy_info`.

PR #271, opened 2025-08-27
Byfield gpu

rohanchanani:
Implements the by field operator for GPUs. To build, configure cmake with -DREALM_USE_CUDART_HIJACK=OFF (necessary) and -DCMAKE_CUDA_ARCHITECTURES={your arch} (to speed up build). To test, run ./tests/deppart basic -ll:gpu 1
@rohany


src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+template <int N, typename T, typename FT>
+void GPUByFieldMicroOp<N,T,FT>::gpu_populate_bitmasks()
+{
+
+    nvtx_range_push("cuda", "byfield_gpu_populate_bitmasks");
+
+    cudaStream_t stream;
+    CUDA_CHECK(cudaStreamCreate(&stream), stream);
+
+    Memory my_mem = field_data[0].inst.get_location();
```
rohany, line 28: Are all the instances gauranteed to be in the same memory? Do you check this anywhere?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+    nvtx_range_push("cuda", "build parent and instance entries");
+
+    std::vector<size_t> inst_offsets(field_data.size() + 1);
```
rohany, line 32: Needs a comment about what this is computing

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+      }
+    }
+    // final end offset
+    inst_offsets[field_data.size()] = inst_size;
+
+    RegionInstance inst_entries_instance = this->realm_malloc(inst_size * sizeof(SparsityMapEntry<N,T>), my_mem);
+    SparsityMapEntry<N,T>* d_inst_entries = reinterpret_cast<SparsityMapEntry<N,T>*>(AffineAccessor<char,1>(inst_entries_instance, 0).base);
+
+    size_t inst_pos = 0;
+    for (size_t i = 0; i < field_data.size(); ++i) {
```
rohany, line 50: I can't tell what's happening here but it seems like there should be a way to create an allocation then turn it into a single cudaMemcpy call.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+      } else {
+        auto tmp = inst_space.sparsity.impl()->get_entries();
+        CUDA_CHECK(cudaMemcpyAsync(d_inst_entries + inst_pos, tmp.data(), tmp.size() * sizeof(SparsityMapEntry<N,T>), cudaMemcpyHostToDevice, stream), stream);
+        inst_pos += tmp.size();
+      }
+    }
+
+    span<SparsityMapEntry<N,T>> parent_entries;
+    std::vector<SparsityMapEntry<N,T>> parent_entries_vec;
+    RegionInstance parent_entries_instance;
```
rohany, line 67: These didn't need to be declared without initialization

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    if (parent_space.dense()) {
+      SparsityMapEntry<N,T> entry;
+      entry.bounds = parent_space.bounds;
+      parent_entries_vec = {entry};
+      parent_entries = span<SparsityMapEntry<N,T>>(parent_entries_vec.data(), 1);
+    } else {
+      parent_entries = parent_space.sparsity.impl()->get_entries();
+    }
+    parent_entries_instance = this->realm_malloc(parent_entries.size() * sizeof(SparsityMapEntry<N,T>), my_mem);
+    d_parent_entries = reinterpret_cast<SparsityMapEntry<N,T>*>(AffineAccessor<char,1>(parent_entries_instance, 0).base);
```
rohany, line 78: You could use these https://github.com/StanfordLegion/realm/blob/main/src/realm/instance.h#L133 methods probably if the AffineAccessor is cumbersome.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    RegionInstance inst_counters_instance = this->realm_malloc((field_data.size()) * sizeof(uint32_t), my_mem);
+    uint32_t* d_inst_counters = reinterpret_cast<uint32_t*>(AffineAccessor<char,1>(inst_counters_instance, 0).base);
+    CUDA_CHECK(cudaMemsetAsync(d_inst_counters, 0, (field_data.size()) * sizeof(uint32_t), stream), stream);
+
+    RegionInstance inst_offsets_instance = this->realm_malloc((field_data.size() + 1) * sizeof(size_t), my_mem);
+    size_t* d_inst_offsets = reinterpret_cast<size_t*>(AffineAccessor<char,1>(inst_offsets_instance, 0).base);
+    CUDA_CHECK(cudaMemcpyAsync(d_inst_offsets, inst_offsets.data(), (field_data.size() + 1) * sizeof(size_t), cudaMemcpyHostToDevice, stream), stream);
+
+    int flattened_threads = 256;
+    int grid_size = (parent_entries.size() * inst_size + flattened_threads - 1) / flattened_threads;
```
rohany, line 93: this intermediate computation could probably overflow an int32 for a large instance.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    intersect_input_rects<N,T><<<grid_size, flattened_threads, 0, stream>>>(d_inst_entries, d_parent_entries, d_inst_offsets, nullptr, inst_size, parent_entries.size(), field_data.size(), d_inst_counters, nullptr);
+    KERNEL_CHECK(stream);
+
+    uint32_t h_inst_counters[field_data.size()+1];
```
rohany, line 97: Why is this randomly a dynamic stack allocation?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    h_inst_counters[0] = 0; // prefix sum starts at 0
+    CUDA_CHECK(cudaMemcpyAsync(h_inst_counters+1, d_inst_counters, field_data.size() * sizeof(uint32_t), cudaMemcpyDeviceToHost, stream), stream);
```
rohany, line 99: Why do you need to copy this back to the host to scan?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    CUDA_CHECK(cudaStreamSynchronize(stream), stream);
+    for (size_t i = 0; i < field_data.size(); ++i) {
+      h_inst_counters[i+1] += h_inst_counters[i];
+    }
+
+    size_t big_inst_counters[field_data.size()+1];
+    for (size_t i = 0; i < field_data.size()+1; ++i) {
```
rohany, line 106: Is this not a memcpy?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    RegionInstance inst_prefix_instance = this->realm_malloc((field_data.size() + 1) * sizeof(size_t), my_mem);
+    size_t* d_inst_prefix = reinterpret_cast<size_t*>(AffineAccessor<char,1>(inst_prefix_instance, 0).base);
+    CUDA_CHECK(cudaMemcpyAsync(d_inst_prefix, big_inst_counters, (field_data.size() + 1) * sizeof(size_t), cudaMemcpyHostToDevice, stream), stream);
+
+    RegionInstance valid_rects_instance = this->realm_malloc(num_valid_rects * sizeof(Rect<N,T>), my_mem);
+    Rect<N,T>* d_valid_rects = reinterpret_cast<Rect<N,T>*>(AffineAccessor<char,1>(valid_rects_instance, 0).base);
+
+    CUDA_CHECK(cudaMemsetAsync(d_inst_counters, 0, (field_data.size()) * sizeof(uint32_t), stream), stream);
+
+    flattened_threads = 256;
```
rohany, line 133: why redefine?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    grid_size = (parent_entries.size() * inst_size + flattened_threads - 1) / flattened_threads;
+    intersect_input_rects<N,T><<<grid_size, flattened_threads, 0, stream>>>(d_inst_entries, d_parent_entries, d_inst_offsets, d_inst_prefix, inst_size, parent_entries.size(), field_data.size(), d_inst_counters, d_valid_rects);
+    KERNEL_CHECK(stream);
+
+    nvtx_range_pop();
+    nvtx_range_push("cuda", "build volume offsets");
+
+    RegionInstance prefix_rects_instance = this->realm_malloc((num_valid_rects + 1) * sizeof(size_t), my_mem);
```
rohany, line 141: Is there any way you can be reusing these instances?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    RegionInstance inst_prefix_instance = this->realm_malloc((field_data.size() + 1) * sizeof(size_t), my_mem);
+    size_t* d_inst_prefix = reinterpret_cast<size_t*>(AffineAccessor<char,1>(inst_prefix_instance, 0).base);
+    CUDA_CHECK(cudaMemcpyAsync(d_inst_prefix, big_inst_counters, (field_data.size() + 1) * sizeof(size_t), cudaMemcpyHostToDevice, stream), stream);
+
+    RegionInstance valid_rects_instance = this->realm_malloc(num_valid_rects * sizeof(Rect<N,T>), my_mem);
+    Rect<N,T>* d_valid_rects = reinterpret_cast<Rect<N,T>*>(AffineAccessor<char,1>(valid_rects_instance, 0).base);
+
+    CUDA_CHECK(cudaMemsetAsync(d_inst_counters, 0, (field_data.size()) * sizeof(uint32_t), stream), stream);
```
rohany, line 131: Does your kernel require this data to be zero initialized?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+
+    RegionInstance accessors_instance = this->realm_malloc(field_data.size() * sizeof(AffineAccessor<FT,N,T>), my_mem);
+    AffineAccessor<FT,N,T>* d_accessors = reinterpret_cast<AffineAccessor<FT,N,T>*>(AffineAccessor<char,1>(accessors_instance, 0).base);
+    CUDA_CHECK(cudaMemcpyAsync(d_accessors, h_accessors, field_data.size() * sizeof(AffineAccessor<FT,N,T>), cudaMemcpyHostToDevice, stream), stream);
+
+    nvtx_range_pop();
+
+    nvtx_range_push("cuda", "populate bitmasks");
+
+    int threads_per_block = 256;
```
rohany, line 217: random redefinition of stuff above?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
+#pragma once
+#include "realm/deppart/byfield.h"
```
rohany, line 2: I've left comments on this file with mostly stylistic things. I'm not yet going to read the other parts of the pipeline, because they share the same problem as here: there are not enough comments or suggestive variable names that describe the overall picture of what this function is actually computing, and how the pieces fit together. Right now, it's not possible to follow the code as someone who didn't write it. This is a pretty algorithmic piece of code -- you should be able to describe in comments (maybe with small examples or descriptions or textual diagrams) about what each phase of the computation is doing, and how that fits into the overall algorithm.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+      d_colors = reinterpret_cast<FT*>(AffineAccessor<char,1>(colors_instance, 0).base);
+      CUDA_CHECK(cudaMemcpyAsync(d_colors, colors.data(), colors.size() * sizeof(FT), cudaMemcpyHostToDevice, stream), stream);
+    }
+
+    AffineAccessor<FT,N,T> h_accessors[field_data.size()];
+    for (size_t i = 0; i < field_data.size(); ++i) {
+      h_accessors[i] = AffineAccessor<FT,N,T>(field_data[i].inst, field_data[i].field_offset);
+    }
+
+    RegionInstance accessors_instance = this->realm_malloc(field_data.size() * sizeof(AffineAccessor<FT,N,T>), my_mem);
```
rohany, line 209: this seems like an example of an instance that could probably live in zero copy memory instead of device memory to avoid a copy.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    nvtx_range_pop();
+
+  std::map<FT, size_t> color_indices;
+  for (size_t i = 0; i < colors.size(); i++) {
+    color_indices[colors[i]] = i;
+  }
+
+  CUDA_CHECK(cudaStreamSynchronize(stream), stream);
+  cudaStreamDestroy(stream);
+  inst_entries_instance.destroy();
```
rohany, line 232: There are instances you deleted here but not in the exit of the fast path. Also, do you need to wait until the end of the function to delete these? You should be trying to free (or reuse) these instances as soon as the data is no longer needed.

rohany (changes requested):
Initial round of feedback

rohany:
Realm has Gtest infrastructure now -- I think you should add your individual tests as unit tests following the patterns here: https://github.com/StanfordLegion/realm/tree/main/tests/unit_tests.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
+#pragma once
```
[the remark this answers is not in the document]
rohanchanani: Should I just make them .h? Or .cuh?

src/realm/deppart/byfield_gpu_impl.hpp
rohanchanani: Yes, these are the counters for the count + emit pass. On the first pass, it counts how many rectangles are to determine how big to alloc. On the second pass, each thread reuses the same counters to figure out where to write. I'll comment.

src/realm/deppart/byfield_gpu_impl.hpp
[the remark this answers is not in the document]
rohany: just .h? be consistent with the style in the repo

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,261 @@
...
+    return r.volume();
+  }
+};
+
+// everything in header form:
+template <int N, typename T, typename FT>
+void GPUByFieldMicroOp<N,T,FT>::gpu_populate_bitmasks()
+{
+
+    nvtx_range_push("cuda", "byfield_gpu_populate_bitmasks");
```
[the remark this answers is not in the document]
rohanchanani: Should I reduce the granularity of my nvtx ranges to avoid vomiting these guards all over the code? i.e. only push and pop at the start and end of each populate function?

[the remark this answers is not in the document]
rohany: Having more detailed ranges is fine, if it helps you. To clean up the code you could probably write a class (add it to the nvtx file) that handles the ranges with RAII so it is simpler to look at. Something like the official nvtx api: https://nvidia.github.io/NVTX/. We can't actually use it directly for packaging reasons, but you could add a utility like this.

src/realm/deppart/partitions.cc
```diff
@@ -71,7 +71,7 @@ namespace Realm {
 				      size_t start, size_t count, size_t volume,
 				      IndexSpace<N,T> *results,
 				      size_t first_result, size_t last_result,
-				      const std::vector<SparsityMapEntry<N,T> >& entries)
+				      const span<SparsityMapEntry<N,T> >& entries)
```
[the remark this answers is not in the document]
rohanchanani: It errors when I do std::span, and elsewhere in realm is just span<> as well - it looks like it's coming from realm/utils.h with a comment that says:
// TODO: actually use C++20 version if available

src/realm/sparsity.h
```diff
@@ -246,8 +246,12 @@ namespace Realm {
 
   protected:
     atomic<bool> entries_valid{false}, approx_valid{false};
-    std::vector<SparsityMapEntry<N, T>> entries;
-    std::vector<Rect<N, T>> approx_rects;
+    std::vector<SparsityMapEntry<N,T> > entries;
+    span<SparsityMapEntry<N,T> > span_entries;
+    std::vector<Rect<N,T> > approx_rects;
+    span<Rect<N,T> > span_approx_rects;
```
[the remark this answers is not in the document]
rohanchanani: Is it better to store instance + size and then a boolean for whether to use instance/vectors than to store the span?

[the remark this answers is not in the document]
rohanchanani: Also, if I construct the span on the fly rather than storing it, should I change the function signature to not return by reference?

tests/deppart.cc
```diff
@@ -161,6 +162,281 @@ int find_split(const std::vector<T> &cuts, T v)
...
+  				assert(p_nodes[i].contains(point.p));
+  			}
+  		}
+  	}
+  	std::cout << "Partitioning correctness check passed" << std::endl;
+  	exit(0);
+        return Event::NO_EVENT;
+  }
+
+  virtual int perform_dynamic_checks(void)
```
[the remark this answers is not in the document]
rohanchanani: The test interface requires definitions for these virtual functions. I moved the checks to check partitioning and left perform_dynamic_checks as a no op.

src/realm/deppart/sparsity_impl.cc
```diff
@@ -457,6 +451,9 @@ namespace Realm {
   template <int N, typename T>
   class SparsityMapToRectAdapter {
   public:
+    SparsityMapToRectAdapter(const span<SparsityMapEntry<N, T>> &_entries)
```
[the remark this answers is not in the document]
rohanchanani: I got rid of it. the original one is used by the cpu sort + merge pipeline, I don't actually need the span overload for the constructor.

src/realm/deppart/partitions_gpu_impl.hpp
```diff
@@ -0,0 +1,269 @@
...
+namespace Realm {
+
+  template<int N, typename T>
+  struct UnionRectOp {
+    __host__ __device__
+    Rect<N,T> operator()(const Rect<N,T>& a,
+                         const Rect<N,T>& b) const {
+      Rect<N,T> r;
+      for(int d=0; d<N; d++){
+        r.lo[d] = a.lo[d] <  b.lo[d] ? a.lo[d] : b.lo[d];
```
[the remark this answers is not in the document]
rohanchanani: nvcc errors when I use std::min and std::max in device code.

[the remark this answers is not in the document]
rohany: What about just min/max? I think nvcc provides stubs for these.

src/realm/deppart/byfield_gpu_impl.hpp
rohany: Yeah, I’m saying that you could probably write it to populate one buffer in host memory and then issue a single memcpy to the device.

src/realm/sparsity.h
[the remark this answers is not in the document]
rohanchanani: Yeah, its size is just a pointer and a size_t. Also, to pass it by reference I'd need to keep it as a member variable.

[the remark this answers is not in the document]
rohany: There’s no need to return a reference then. References are either for mutable access, or a const reference to get a pointer to large data without copying.

src/realm/deppart/byfield.cc
```diff
@@ -322,21 +373,29 @@ namespace Realm {
...
-      uop->dispatch(this, true /* ok to run in this thread */);
+  template<int N, typename T, typename FT>
+  void ByFieldOperation<N, T, FT>::execute(void) {
+
+    if (field_data.size() > 0 && field_data[0].inst.get_location().kind()==Memory::GPU_FB_MEM) {
+      GPUByFieldMicroOp<N, T, FT> *uop = new GPUByFieldMicroOp<N, T, FT>(parent, field_data);
+      for (size_t i = 0; i < colors.size(); i++) {
+        uop->add_sparsity_output(colors[i], subspaces[i]);
+      }
+      uop->dispatch(this, true);
```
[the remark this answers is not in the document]
rohany: I think so? I'm not sure what thread this code actually runs in... If it runs on a background worker then I think we're OK to stay on this thread, but otherwise we shouldn't.

.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ concurrency:
 
 jobs:
   build:
```
rohany, line 31: What is this change?

src/realm/deppart/byfield_gpu_impl.hpp
rohanchanani: It's usually tiny (the number of instances) and I need the last value on the host anyways.

.github/workflows/ci.yml
rohanchanani: I think from a commit I fetched in between.

src/realm/sparsity.h
```diff
@@ -246,8 +246,12 @@ namespace Realm {
...
+    RegionInstance entries_instance;
```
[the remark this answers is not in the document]
rohanchanani: If I define the sparsity map impl destructor, it gets called twice - once I think from going out of scope, and the other from the wrapper calling map deleter (which just does delete impl). For now, I added a guard in the destructor to only destroy the instances the first time, but I could also define a pseudoconstructor that deletes the instance and add a call to that to map deleter.

src/realm/deppart/byfield_gpu_impl.hpp
rohany: I still think this could be 1 cudaMemcpy.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+    RegionInstance parent_entries_instance = this->realm_malloc(parent_entries.size() * sizeof(SparsityMapEntry<N,T>), my_mem);
+
+    RegionInstance inst_counters_instance = this->realm_malloc((field_data.size()) * sizeof(uint32_t), my_mem);
+
+    RegionInstance inst_offsets_instance = this->realm_malloc((field_data.size() + 1) * sizeof(size_t), my_mem);
+
+    SparsityMapEntry<N, T>* d_parent_entries = reinterpret_cast<SparsityMapEntry<N,T>*>(AffineAccessor<char,1>(parent_entries_instance, 0).base);
+    CUDA_CHECK(cudaMemcpyAsync(d_parent_entries, parent_entries.data(), parent_entries.size() * sizeof(SparsityMapEntry<N,T>), cudaMemcpyHostToDevice, stream), stream);
+
+    //This is used for count + emit: first pass counts how many rectangles survive intersection, second pass uses the counter
```
rohany, line 96: Isnt this producing points, not rectangles?

src/realm/deppart/byfield_gpu_impl.hpp
rohany: `int` by default is not necessarily int64_t, so this could still overflow. Also, it is common to extract this logic into a macro or function since it will be used repeatedly.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    size_t num_valid_rects = h_inst_counters[field_data.size()];
+
+    if (num_valid_rects == 0) {
+      for (std::pair<const FT, SparsityMap<N, T>> it : sparsity_outputs) {
+        SparsityMapImpl<N, T> *impl = SparsityMapImpl<N, T>::lookup(it.second);
+        impl->gpu_finalize();
+      }
+      inst_offsets_instance.destroy();
```
rohany, line 130: I don't see some of these used in the `else` case until they are deleted later anyway. Can they just be unconditionally deleted?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+                    >;
+    VolIter d_volumes(d_valid_rects, RectFieldVolumeOp<N,T>());
+
+    void*   d_temp = nullptr;
+    size_t rect_temp_bytes = 0;
+    cub::DeviceScan::InclusiveSum(
+        /* d_temp_storage */  nullptr,
+        /* temp_bytes */      rect_temp_bytes,
+        /* d_in */            d_volumes,
+        /* d_out */           d_prefix_rects + 1,   // shift by one so prefix[1]..prefix[n]
```
rohany, line 176: Why are you shifting by one then doing an inclusive scan? Isn't that the same as a normal exclusive scan?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+        /* num_items */       num_valid_rects, stream);
+
+    RegionInstance temp_instance = this->realm_malloc(rect_temp_bytes, my_mem);
```
rohany, line 179: What is this second scan + temporary?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+    inst_counters_instance.destroy();
+
+    RegionInstance points_instance = this->realm_malloc(total_pts * sizeof(PointDesc<N,T>), my_mem);
+    PointDesc<N,T>* d_points = reinterpret_cast<PointDesc<N,T>*>(AffineAccessor<char,1>(points_instance, 0).base);
+
+    FT* d_colors;
+    RegionInstance colors_instance;
+
+    //Memcpying a boolean vector breaks things for some reason so we have this disgusting workaround
+    if constexpr(std::is_same_v<FT,bool>) {
```
rohany, line 207: Is making `FT` a bool even allowed?

rohany: The reason for this though is that `vector<bool>` is specialized to an implementation that packs 8 elements into a single byte, rather than a byte per element.

src/realm/deppart/byfield_gpu_impl.hpp
rohany: Bumping this, could just be a zero copy vector of accessors.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    RegionInstance accessors_instance = this->realm_malloc(field_data.size() * sizeof(AffineAccessor<FT,N,T>), my_mem);
+    AffineAccessor<FT,N,T>* d_accessors = reinterpret_cast<AffineAccessor<FT,N,T>*>(AffineAccessor<char,1>(accessors_instance, 0).base);
+    CUDA_CHECK(cudaMemcpyAsync(d_accessors, h_accessors.data(), field_data.size() * sizeof(AffineAccessor<FT,N,T>), cudaMemcpyHostToDevice, stream), stream);
+
+
+    //This is where the work is actually done - each thread figures out which points to read, reads it, marks a PointDesc with its color, and writes it out
+    int num_blocks = (total_pts + flattened_threads - 1) / flattened_threads;
+    byfield_gpuPopulateBitmasksKernel<N,T,FT><<<num_blocks, flattened_threads, 0, stream>>>(d_accessors, d_valid_rects, d_prefix_rects, d_inst_prefix, d_colors, total_pts, colors.size(), num_valid_rects, field_data.size(), d_points);
+    KERNEL_CHECK(stream);
```
rohany, line 238: Do you need separate `KERNEL_CHECK` and `CUDA_CHECK` macros?

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+    CUDA_CHECK(cudaMemcpyAsync(&total_pts, &d_prefix_rects[num_valid_rects], sizeof(size_t), cudaMemcpyDeviceToHost, stream), stream);
+
+    CUDA_CHECK(cudaStreamSynchronize(stream), stream);
+    inst_entries_instance.destroy();
+    parent_entries_instance.destroy();
+    inst_offsets_instance.destroy();
+    temp_instance.destroy();
+    inst_counters_instance.destroy();
+
+    RegionInstance points_instance = this->realm_malloc(total_pts * sizeof(PointDesc<N,T>), my_mem);
```
rohany, line 200: Stylistically, i would declare this variable closer to where it is used for the first time.

src/realm/deppart/byfield_gpu_impl.hpp
```diff
@@ -0,0 +1,268 @@
...
+      if (field_data[i].index_space.dense()) {
+        inst_size += 1;
+      } else {
+        inst_size += field_data[i].index_space.sparsity.impl()->get_entries().size();
+      }
+    }
+    inst_offsets[field_data.size()] = inst_size;
+
+
+    RegionInstance inst_entries_instance = this->realm_malloc(inst_size * sizeof(SparsityMapEntry<N,T>), my_mem);
```
rohany, line 56: If I remember the presentation correctly, could this first group of work (intersecting source space rectangles etc) be extracted into a separate helper that the other ops will use too?

src/realm/deppart/byfield_gpu_kernels.hpp
```diff
@@ -0,0 +1,110 @@
...
+  assert(idx_y < numInstRects);
+  const SparsityMapEntry<N, T> source_entry = d_parent_entries[idx_x];
+  const SparsityMapEntry<N, T> inst_entry = d_inst_entries[idx_y];
+  Rect<N,T> rect_output = inst_entry.bounds.intersection(source_entry.bounds);
+  if (rect_output.empty()) {
+    return;
+  }
+  size_t low = 0, high = numInsts;
+  while (low < high) {
+      size_t mid = (low + high) >> 1;
```
rohany, line 35: look up and do the "no overflow" binary search. This can overflow if low and high are too big. It's unlikely to happen here because numInsts is small but just to use the "correct" algorithm.

src/realm/deppart/byfield_gpu_kernels.hpp
```diff
@@ -0,0 +1,110 @@
...
+      if (d_inst_offsets[mid+1] <= idx_y) low = mid + 1;
+      else                                 high = mid;
+  }
+  size_t inst_idx = low;
+  uint32_t local = atomicAdd(&d_inst_counters[inst_idx], 1);
```
rohany, line 40: When `d_rects` is not null, do you need to do the atomic add? It doesn't seem that way in the code above.

src/realm/deppart/byfield_gpu_kernels.hpp
```diff
@@ -0,0 +1,110 @@
...
+  while (low < high) {
+    size_t mid = (low + high) >> 1;
+    if (prefix[mid+1] <= idx) low = mid + 1;
+    else                      high = mid;
+  }
+  size_t r = low;
+
+  // Binary search to find which instance this rectangle belongs to
+  low = 0, high = num_insts;
+  while (low < high) {
```
rohany, line 79: This is binary search #3, probably worth it to put it into a helper function now for clarity.

src/realm/deppart/partitions_gpu_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+  } while (0)
+
+
+//NVTX macros to only add ranges if defined
+#ifdef REALM_USE_NVTX
+
+  #define NVTX_CAT(a,b)  a##b
+
+  #define NVTX_DEPPART(message) \
+  nvtxScopedRange NVTX_CAT(nvtx_, message)("cuda", #message, 0)
```
rohany, line 41: shouldn't the message be `deppart` or something?

src/realm/deppart/partitions_gpu_impl.hpp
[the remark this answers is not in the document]
rohany: Did you undo this change?

src/realm/deppart/partitions_gpu_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+    RegionInstance break_points_instance = this->realm_malloc(total_pts * sizeof(uint8_t), my_mem);
+
+    const uintptr_t aux_in = AffineAccessor<char,1>(aux_instance, 0).base;
+    const uintptr_t aux_out = aux_in + max_aux_bytes;
+
+    T* d_keys_in = reinterpret_cast<T*>(aux_in);
+    T* d_keys_out = reinterpret_cast<T*>(aux_out);
+
+    PointDesc<N,T>* d_points_in = d_points;
+    PointDesc<N,T>* d_points_out = reinterpret_cast<PointDesc<N,T>*>(AffineAccessor<char,1>(pg_instance, 0).base);
```
rohany, line 108: Why are you doing reinterpret cases instead of making the affine accessor with the right type?

src/realm/deppart/partitions_gpu_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+    size_t* group_ids = reinterpret_cast<size_t*>(AffineAccessor<char,1>(pg_instance, 0).base);
+
+    RectDesc<N,T>* d_rects_in = reinterpret_cast<RectDesc<N,T>*>(aux_in);
+    RectDesc<N, T> *d_rects_out = reinterpret_cast<RectDesc<N,T>*>(aux_out);
+
+    size_t* d_src_keys_in = reinterpret_cast<size_t*>(aux_in);
+    size_t* d_src_keys_out = reinterpret_cast<size_t*>(aux_out);
+
+    size_t t1=0, t2=0, t3=0;
+    cub::DeviceRadixSort::SortPairs(nullptr, t1, d_keys_in, d_keys_out, d_points_in, d_points_out, total_pts, 0, 8*sizeof(T), stream);
```
rohany, line 121: what are these 8*T sizes?

src/realm/deppart/partitions_gpu_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+    cub::DeviceScan::InclusiveSum(nullptr, t3, break_points, group_ids, total_pts, stream);
+
+    //Temporary storage instance shared by CUB operations.
+    size_t temp_bytes = std::max({t1, t2, t3});
+    RegionInstance temp_storage_instance = this->realm_malloc(temp_bytes, my_mem);
+    void *temp_storage = reinterpret_cast<void*>(AffineAccessor<char,1>(temp_storage_instance, 0).base);
+
+
+    //Sort along each dimension from LSB to MSB (0 to N-1)
+    int threads_per_block = 256;
```
rohany, line 132: yep, pull this into a macro near the other macros.

src/realm/deppart/partitions_gpu_impl.hpp
```diff
@@ -0,0 +1,343 @@
...
+  /*
+   *  Input: An array of disjoint rectangles sorted by src idx.
+   *  Output: Fills the sparsity output for each src with a host region instance
+   *  containing the entries/approx entries and calls gpu_finalize on the SparsityMapImpl.
+   *  Approach: Segments the rectangles by their src idx and copies them back to the host,
+   */
+
+  template<int N, typename T>
+  template<typename Container, typename IndexFn, typename MapFn>
+  void GPUMicroOp<N,T>::send_output(RectDesc<N, T>* d_rects, size_t total_rects, Memory my_mem, const Container& ctr, IndexFn getIndex, MapFn getMap)
```
rohany, line 208: Let's add this part of the pipeline in a future PR when we need it.

src/realm/deppart/sparsity_impl.cc
```diff
@@ -907,6 +902,19 @@ namespace Realm {
     , sparsity_comm(_sparsity_comm)
   {}
 
+template<int N, typename T>
+SparsityMapImpl<N, T>::~SparsityMapImpl(void)
+{
+     //We are responsible for our instances
+     if (this->from_gpu) {
+       if (get_runtime()->get_memory_impl(this->entries_instance))
```
rohany, line 910: What is a case where the memory no longer exists but the instance does?

src/realm/deppart/sparsity_impl.cc
```diff
@@ -907,6 +902,19 @@ namespace Realm {
...
+         this->entries_instance.destroy();
+       if (get_runtime()->get_memory_impl(this->approx_instance))
+         this->approx_instance.destroy();
+       this->from_gpu = false;
```
rohany, line 914: Why do you need to set `from_gpu` to be false here? The object is being destructed. Also, Why do you need a `from_gpu` bit in the first place? Isn't that apparent from the existence of the isntances?

src/realm/deppart/sparsity_impl.cc
```diff
@@ -1494,17 +1501,16 @@ namespace Realm {
         assert(false);
       // scan the entry list, sending bitmaps first and making a list of rects
       std::vector<Rect<N, T>> rects;
-      for(typename std::vector<SparsityMapEntry<N, T>>::const_iterator it =
-              this->entries.begin();
-          it != this->entries.end(); it++) {
-        if(it->bitmap) {
+      for(size_t i = 0; i < this->get_entries().size(); i++) {
+        SparsityMapEntry<N, T> entry = this->get_entries()[i];
```
rohany, line 1505: probably `const SparsityMapEntry<N, T>& entry` here.

src/realm/deppart/sparsity_impl.cc
```diff
@@ -1830,6 +1839,117 @@ namespace Realm {
...
+      GenEventImpl::trigger(trigger_approx, false /*!poisoned*/);
+
+    if(trigger_precise.exists())
+      GenEventImpl::trigger(trigger_precise, false /*!poisoned*/);
+  }
+
+
+  //Allows a GPU deppart client to set the entries directly with a host region instance
+  template<int N, typename T>
+    void SparsityMapImpl<N, T>::set_instance(RegionInstance _entries_instance, size_t size)
```
rohany, line 1941: If we're always using `set_instance` from the gpu deppart ops, why do we need to copy/expose a separate `gpu_finalize` function?

src/realm/indexspace.inl
```diff
@@ -591,41 +591,38 @@ namespace Realm {
...
-        if(it->sparsity.exists()) {
-          assert(0);
-        } else if(it->bitmap != 0) {
-          assert(0);
-        } else {
-          return true;
-        }
+      for(size_t i = 0; i < entries.size(); i++) {
+        SparsityMapEntry<N, T> entry = entries[i];
+	if(!entry.bounds.contains(p)) continue;
```
rohany, line 618: odd formatting

rohany (changes requested):

src/realm/deppart/byfield_gpu_impl.hpp
rohanchanani: Cub calls need two runs, the first is to figure out how big your temporary should be and the second actually does the comptuation.

src/realm/deppart/byfield_gpu_kernels.hpp
rohanchanani: Yes because I prefix sum the counters do figure out how big to make the output and to know where to write each output rectangle in the emit.

src/realm/deppart/partitions_gpu_impl.hpp
rohanchanani: Byfield uses send_output - it's called at the end of complete_pipeline.

rohanchanani:
@rohany @lightsighter @artempriakhin Here's a writeup for the two proposals on how to handle dynamic allocations:

https://docs.google.com/presentation/d/1JQrFJlsi_BNijg2eTmTd3DxqYCBQYjiZsqNl0FP8k10/edit?usp=sharing

Proposal 1: If an allocation fails/is deferred, recycle all the sparsity map outputs and set their ids to 0. The user is then responsible for checking the sparsity IDs of their output index spaces once the partitioning op event resolves and handling failure.
Proposal 2: Break each operation into multiple calls, with allocations between the calls based on calculated sizes. The total # of required functions for a single partitioning op range from 3 to 7, although many of the functions can be used for multiple ops.

lightsighter:
Rather than respond to each of the comments above for each different proposal, I'm going to focus on this:

> > Proposal 5: (from @streichler) reserve a small amount of GPU memory for all such dependent partitioning operations, optimistically try to allocate buffers (e.g. in the style of Proposal 1) and then fall back to a more serialized execution using the pre-allocated small buffer as necessary. Adding my own idea to this one: the buffer could even be in zero-copy or UVM memory in order to avoid unnecessarily consuming device-side memory.

> I think this makes a lot of sense. One consideration is that the lifetime of GPU deppart operations will generally be a small fraction of the application's runtime, but they need a meaningful chunk of memory for their small burst - maybe this could include a way for the client to reclaim the memory when they know they aren't doing any partitioning?

If it is possible for the code to potentially work out of a smaller buffer as well as running faster with a larger buffer, then I think the interface I would like would be a two-phase model. In the first phase we do some kind of API call that estimates bounds on memory usage and returns two sizes: a minimal size needed to run the computation and an "optimal" size that ensures that deppart can get "good" performance (I feel like this value should be more a function of the GPU being targeted and not a property of the data being used for the deppart operation, but tell me if I am wrong about that assumption). If Realm is going to run on CPUs it will just return zero for both values (or maybe a small value, not sure if we use buffers on the CPU implementation). The client can then do an allocation at either value (or somewhere in-between if they're willing to search) and pass that in to the implementation. 

Tell me if that is possible and if not then we'll go back to exploring the other proposals.

> Legion is by far the biggest consumer of Realm, so I'll gladly defer to whatever combination of these proposals you think would be the most natural/intuitive for incorporating GPU-deppart into Legion while maintaining Realm's invariants.

Even though I care a lot about the Legion implementation, we've always designed Realm in a way so that it is Legion-agnostic. Realm should be usable by lots of different clients (I'm actually working on a different project built on Realm right now, it's just not public) so we try very hard to maintain that independence when designing APIs for Realm.

lightsighter:
> Say you want to do the computation with memory size K such that K < "optimal" size.

Can you say what your definition of the "optimal size" is? To be fair, let me give you my definition of "optimal size". In general, all of these dependent partitioning operations are going to be memory bandwidth limited. You can saturate the memory bandwidth of a GPU with many fewer SMs than most GPUs have (only really tiny GPUs are balanced and you'll mostly find them in laptops). Therefore, my definition of "optimal size" is however much memory I would need to allocate in order to have a big enough working set for the number of SMs required to saturate memory bandwidth. After that, everything else should just be executed in serial because even if I "parallelize" it with more blocks/threads, it's also going to be implicitly serialized anyway by the memory controller for the HBM memory. 

> Any accommodation I can think of to these challenges ends up serializing things to the point of defeating the purpose of gpu partitioning

If you serialize things only after having saturated memory bandwidth, then you're not going to lose any performance.

lightsighter:
To further elaborate, you can do all this with a single kernel launch with a small number of threadblocks and using a [cooperative kernel](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EXECUTION.html#group__CUDART__EXECUTION_1g504b94170f83285c71031be6d5d15f73) and then the GPU version of the algorithm a bit like the CPU algorithm with the inner loop(s) parallelized across some number of threads/threadblocks with necessary synchronization inserted using [cooperative groups](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#cooperative-groups). You can then scale it up or down as necessary depending on the available memory bandwidth of the particular GPU you're running on. Bonus points if you can make it use temporal locality to leverage the L2 cache.

rohanchanani:
What I mean by “optimal” size is enough to keep all the data the partitioning operation needs on the device for the entire operation. 

> Say you want to do the computation with memory size K such that K < "optimal" size.

What I mean here is that in order to to use any data less than the “optimal” size, you have to tile the operation in some way. 

> If you serialize things only after having saturated memory bandwidth, then you're not going to lose any performance.

In the approach you’re describing, the tilling would be such that the total amount of work done stays the same, but the active working set at any given time is less. Thus, there’d only be performance degradation if you’re not saturating memory bandwidth, but no matter what the total amount of computation you’re doing would be the same (just N pieces of size 1 / N rather than 1 whole piece).  

I don’t think it’s possible to tile the operations in this way because of the last step, where all the points/rectangles are coalesced into a sparsity map, because the points/rectangles aren’t independent (they need to know if they duplicate/overlap any others). If you split the operation into N pieces, you’d have to do the following: coalesce the first 1/N points, keep that active somewhere, coalesce the next 1/N points, merge that into the working set (more expensive than the coalescing), coalesce the next 1/N points, merge that, etc. So rather than doing the same amount of work with a smaller working set, this adds a meaningful amount of computation that wouldn’t have originally been done with 1 big piece (all the merge operations). Each merge operation would have to do the slide 19 algorithm on the combination of the two spaces: https://docs.google.com/presentation/d/1Iwo0IwXBk14-E8i5i5kLN2B0KP7Cl-RwC0yDIN0_Fh8/edit?usp=sharing

> Any accommodation I can think of to these challenges ends up serializing things to the point of defeating the purpose of gpu partitioning, although I'm definitely open to if there's a better way to do things in less than the "optimal" size.

Here I don’t mean serialized in terms of less block/thread parallelism, but rather in terms of 1 bulk operation serialized into N tile operations, each of which requires a new, expensive integration step in between. And the nature of the computation goes from something that’s well-suited for the GPU (dumping a bulk operation onto the device and treating its elements as independent until the last possible moment) to something that’s generally better suited for the CPU (maintaining an active, disjoint state and iteratively merging pieces into it, which is also exactly what the CPU deppart path currently does). 

> But to know the memory used at stage N for chunk C, you need to have already computed each stage n < N.

I also still think this is a key challenge for any tilling approach with a fixed size buffer. Given a fixed size you have to stay within, I don’t think it’s possible to efficiently choose an input chunk size that's guaranteed to stay within the given size for the entire operation, i.e. to efficiently invert the pipeline. So when you choose a tile, there's a chance that tile will get to say the third stage and increase to bigger than your buffer, at which point the operation would have to fail.

If it's helpful, we could also hop on a call to discuss this.

rohany:
> Presuming you're keeping the rectangle data in shared memory, you can even have all the threads looking for conflicts by going through the [distributed shared memory](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#distributed-shared-memory) network across blocks so you won't incur any extra global memory traffic.

I don't think that it's appropriate to try and use DSMEM here, as only 2 "adjacent" CTAs can access each others' shared memory.

Taking a step back, I want to understand the root of your concerns here. A hybrid algorithm like the one you describe may be possible (@rohanchanani will be the arbiter of that), but what's the main reason that you are pushing for it? Is it mostly for avoiding the complexity of adding the multi-stage deppart API in Realm, or because you think that we should just doing better with the algorithmic components themselves? I am a bit worried about the complexity and cpu-side overheads (launching lots of kernels and copies) of doing a hybrid algorithm with a small tile size -- if it's just performance we can use green contexts or something to limit the number of SM's and allow other application work to proceed. If it is to avoid the multi-stage API in Realm, perhaps @rohanchanani can try to develop an algorithm for one of the deppart operations that does work in a tiled manner and see what the relative performance and code complexity tradeoffs are.

It also seems like if there is even a single deppart operation that we wouldn't be able to apply this to (like maybe the dynamic allocations for the BVH in preimage can't be made to fit in a fixed buffer) and we have to add a staged API then we might as well add it for all deppart operations.

lightsighter:
> what's the main reason that you are pushing for it?

The only way for Legion to implement an unbounded memory requirement operation is to effectively map an unbounded pool in that memory which blocks all downstream operations from allocation/deallocating in that memory. I don't want that to be the default for GPU-accelerated deppart operations in Legion. Any amount of complexity is worth it to avoid that as the effects ripple far up the software stack (beyond Legion). I could care less about the implementation complexity in Legion, but as soon as the consequences become visible to Legion users then I care A LOT and am willing to go to any ends to avoid it if possible.

> if it's just performance we can use green contexts or something to limit the number of SM's and allow other application work to proceed. If it is to avoid the multi-stage API in Realm, perhaps @rohanchanani can try to develop an algorithm for one of the deppart operations that does work in a tiled manner and see what the relative performance and code complexity tradeoffs are

It's neither. It's the fact that the memory requirements ripple up the software stack and are visible to mappers and impact unrelated operations which is something that is not currently the case for deppart operations.

> In particular, if we need tiles in the 10%+ capacity of the GPU to actually get reasonable performance out of this pathway then that already feels like too much to carve out into a fixed pool and we'd want to not have the application be forced to make this choice a-priori.

You should be able to do this with way less than 10% of memory. I bet having a buffer on the order of 10-100MB is enough.

> It also seems like if there is even a single deppart operation that we wouldn't be able to apply this to (like maybe the dynamic allocations for the BVH in preimage can't be made to fit in a fixed buffer) and we have to add a staged API then we might as well add it for all deppart operations.

Also strong disagree here. If we have to let consequences ripple up the software stack, then we should tightly bound the kinds of operations that it applies to.

rohanchanani:
I can definitely get started on a tiled approach for the point-based pipelines - also, preimage range dumps into the points pipeline, so of the non-setops only image range requires rects. But I also think that the back-end of tiling will end up looking pretty close to identical for both points and rects because we'll want to store the active set as rectangles.

lightsighter:
> I finished a tiled implementation of image that should also work for the other operations. 

Congrats! Very excited about this!

> I will be able to give a "minimal size" to complete the operation and an "optimal size"

Three questions about this:
1. Will the "optimal" size be guaranteed to do the whole operation in one "tile" or is it just a best guess?
2. Did you propose an API for querying this somewhere? I didn't immediately see it. Might be easier to spot it if you merged from `main` so there aren't so many outstanding diffs.
3. If the user picks a buffer size, can they just linearly interpolate how many tiles will be needed, or will we need to measure that at runtime and report it in a profiling response from the deppart operation?

> although it seems the optimal size will be based on what it takes to do the operation in one go rather than what it takes to saturate memory bandwidth as the cost of tilling is significant.

Two questions about this:
1. Can you elaborate on the sources of the failures? I missed the first half of the talk last week. I can see how we might do redundant work in deduplicating rectangles/points in constructing the final sparsity map, but I'm not clear on how things might fail in the middle.
4. Can you plot (or just report) the "effective memory bandwidth" that is achieved even with the wasted/redundant work? It would be good to confirm that we're actually hitting the memory bandwidth limit in all cases, even if the result is that some (a lot) of that bandwidth is going to waste. Maybe we discover we're actually compute bound (a pleasant surprise although doubtful) or that there is some other limiting factor (e.g. branching/memory divergence) which we might be able to improve in some way (e.g. sorting the way we do the work).

> I'm also working on a host fallback for when the output itself doesn't fit in the device buffer, but the code above is the primary consideration for feedback/design decisions.

Seems reasonable. This path just needs to be functional and not necessarily fast.

PR #272, opened 2025-08-27
Instance Redicstricting for Dynamic Frambuffer Memory

lightsighter:
This branch adds supports for instance redistricting on instances allocated in the dynamic GPU framebuffer memory. It does this by keeping the underlying CUDA allocation alive as long as there is any instance alive that still refers to the underlying allocation. Once all the instances redistricted on top of the base allocation have been destroyed then the underlying CUDA allocation is freed back to the CUDA driver. This unfortunately means that there is a potential for fragmentation as a large initial allocation could result in just a few small instances surviving after redistricting, but it is at least functionally correct.


apryakhin:
@lightsighter is this needs to be reviewed now or we are in POC/testing stage?

PR #286, opened 2025-09-03
Add hash functions to id and some POD classes.

### LINKED ISSUE #114: Add `std::hash` specialization for Realm objects
For example:
- `Realm::Memory`

muraj:
Closes #114.

### DIFF
--- a/src/realm/event.inl
+++ b/src/realm/event.inl
@@ -19,6 +19,7 @@
 
 // nop, but helps IDEs
 #include "realm/event.h"
+#include <functional>
 
 #include "realm/serialize.h"
 TYPE_IS_SERIALIZABLE(Realm::Event);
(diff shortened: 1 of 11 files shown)


src/realm/processor.inl
```diff
@@ -16,6 +16,7 @@
  */
 
 // processors for Realm
+#include <functional>
```
Jacobfaib, line 19: Note, you technically don't need to include `<functional>` (which is among the most heavy-weight includes in the STL) for `std::hash`. Any one of the headers listed in https://en.cppreference.com/w/cpp/utility/hash.html will include it.

src/realm/point.h
```diff
@@ -141,6 +141,7 @@ namespace Realm {
 
   template <int N, typename T>
   struct REALM_PUBLIC_API Rect {
+    typedef ONLY_IF_INTEGRAL(T) value_type;
```
Jacobfaib, line 144: Why `typedef` instead of `using`? Or is that for another day?

eddy16112 (commented):
I am wondering what is the use case of having a hash for all these classes?

src/realm/point.h
muraj: Another day maybe.

src/realm/event.inl
```diff
@@ -103,3 +104,24 @@ namespace Realm {
   }
 
 }; // namespace Realm
+
+template<>
+struct std::hash<Realm::Event> {
```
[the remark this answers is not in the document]
muraj: > Can we use REALM_HASH_DEFINE(Realm::Event), such that we do not need to write it for each class?

I would need to define this in a common header across all of these.  I don't really think it's worth it, this should be fine for these few cases.

src/realm/processor.inl
Jacobfaib: It's a header, so not just your builds are affected, but ours (and any of our downstream) as well. Idk, it's one of those things that tends to creep until suddenly it's a huge problem.

muraj: And we can deal with it then.  Again, functional is a common header we include in a lot of other places, you're probably including it already, so it doesn't make much of a difference here.

PR #296, opened 2025-09-11
Restore previous altstack on thread exit

Jacobfaib:
Instead of disabling the sigaltstack, we should restore the previous one. Libraries like AddressSanitizer for example will install a sigaltstack of its own, which we should restore before killing the thread.

### DIFF
--- a/src/realm/cuda/cuda_memcpy.cu
+++ b/src/realm/cuda/cuda_memcpy.cu
@@ -66,65 +66,70 @@ static __device__ inline size_t coords_to_index(Offset_t *coords, const Offset_t
 }
 
 template <typename T, typename Offset_t = size_t>
-static __device__ inline void memcpy_kernel_transpose(
-    Realm::Cuda::MemcpyTransposeInfo<Offset_t> info, T* tile) {
+static __device__ inline void
+memcpy_kernel_transpose(Realm::Cuda::MemcpyTransposeInfo<Offset_t> info, T *tile)
+{
   __restrict__ T *out_base = reinterpret_cast<T *>(info.dst);
   __restrict__ T *in_base = reinterpret_cast<T *>(info.src);
   const Offset_t tile_size = info.tile_size;
+
   const Offset_t tidx = threadIdx.x % tile_size;
   const Offset_t tidy = (threadIdx.x / tile_size) % tile_size;
+
   const Offset_t grid_dimx = ((info.extents[2] + tile_size - 1) / tile_size);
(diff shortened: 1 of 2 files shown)


src/realm/threads.cc
```diff
@@ -873,29 +875,24 @@ namespace Realm {
...
-      //  up properly, so our stack may not have been active anyway
-      // either way though, it's not active after the call above, so it's
-      //  safe to free the memory now
-      // assert(oldstack.ss_sp == thread->altstack_base);
+      if(current_stack.ss_sp == thread->altstack_base) {
+        // The current stack is the one we installed previously. It is safe to restore the
+        // old one.
+        ret = sigaltstack(&old_stack, nullptr);
+        assert(ret == 0);
+      } else {
```
muraj, line 888: Please don't add empty else cases.  Just merge the else comment with the one in the if, or move both comments to just before the if.

src/realm/threads.cc
```diff
@@ -843,13 +843,15 @@ namespace Realm {
...
+    stack_t old_stack{};
+
     // install our alt stack (if it exists) for signal handling
     if(thread->altstack_base != 0) {
       stack_t altstack;
       altstack.ss_sp = thread->altstack_base;
       altstack.ss_flags = 0;
       altstack.ss_size = thread->altstack_size;
-      int ret = sigaltstack(&altstack, 0);
+      int ret = sigaltstack(&altstack, &old_stack);
```
muraj, line 854: Will old_stack here have the disabled flag if no sigaltstack was called before this one?  If not, then shouldn't we call the following sigalstack later on with the oldstack but with the disabled flag set?

Jacobfaib: On macOS and Ubuntu 22.04 it contains `SS_DISABLE`:
```
#include <csignal>
#include <iostream>

int main()
{
  stack_t old_stack;

  sigaltstack(nullptr, &old_stack);
  std::cout << "ptr " << old_stack.ss_sp << "\n";
  std::cout << "size " << old_stack.ss_size << "\n";
  std::cout << "flags " << old_stack.ss_flags << "\n";

  std::cout << "SS_ONSTACK " << SS_ONSTACK << "\n";
  std::cout << "SA_DISABLE " << SS_DISABLE << "\n";
}

$ clang++ stack.cpp -o stack && ./stack
ptr 0x0
size 0
flags 4
SS_ONSTACK 1
SA_DISABLE 4
```

PR #300, opened 2025-09-15
inst: Add a way to get an event when the destruction of an instance actually occurs

muraj:


muraj:
@lightsighter updated to always return an event.  It'll not create a destroy event if wait_on is Event::NO_EVENT, and just return Event::NO_EVENT in that case, which is always triggered, as a small optimization.  I can go deeper and only do this when the wait_on event is triggered, but I'm not sure it's worth it.

muraj:
> if I issue the destroy on a remote node

Ah, hm, no, I'll have to add that.

PR #311, opened 2025-09-22
Signal handler for SIGTERM

suranap:
This change introduces a signal handler for SIGTERM. When the Legion application receives a SIGTERM, this handler will:
1. Call Realm::LoggerConfig::flush_all_streams() to ensure all buffered log messages from both Realm and Legion loggers are written to disk.
2. Log a message to stderr indicating that SIGTERM was caught.
3. Reset the SIGTERM handler to the default disposition.
4. Re-raise SIGTERM to allow the application to terminate as expected.

This ensures that logs are not lost if the application is terminated via SIGTERM. The handler is registered in RuntimeImpl::finish_configure, alongside other existing signal handlers.

Add comprehensive test for SIGTERM signal handling

This test verifies the SIGTERM handler functionality added in previous commits:
- Forks a child process running a Realm application with long-running tasks
- Sends SIGTERM to the child process after startup
- Verifies the process terminates correctly with SIGTERM signal
- Checks for proper log flushing behavior during signal handling

The test confirms that Realm properly catches SIGTERM, flushes all buffered logs, and allows graceful termination without losing log data.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -448,6 +448,10 @@ target_link_libraries(c_external_inst ${TEST_GPU_LIBS})
 
 add_integration_test(c_event_poisoned "${REALM_TEST_DIR}/c/test_event_poisoned.cc")
 add_integration_test(sigterm_test "${REALM_TEST_DIR}/sigterm_test.cc")
+# SIGTERM handling test uses POSIX signals not available on Windows
+if(WIN32)
+  set_property(TEST sigterm_test PROPERTY DISABLED TRUE)
+endif()
 
 target_link_libraries(
   machine_config_test ${TEST_GPU_LIBS} $<TARGET_NAME_IF_EXISTS:CUDA::cuda_driver>


src/realm/runtime_impl.cc
```diff
@@ -373,6 +374,33 @@ namespace Realm {
...
+    sigaction(SIGTERM, &sa_ign, NULL);
+
+    fprintf(stderr, "Realm caught SIGTERM, flushing logs...\n");
+    fflush(stderr);
+    Realm::flush_all_streams();
+    fprintf(stderr, "Realm logs flushed. Re-raising SIGTERM.\n");
+    fflush(stderr);
+
+    // Reset to default and re-raise
+    sigaction(SIGTERM, &sa_dfl, NULL);
```
muraj, line 400: We don't want to reset to default in case a SIGTERM handler was specified before initializing realm.  We need to stash the original handler and call it as needed.

src/realm/runtime_impl.cc
```diff
@@ -1992,6 +2020,15 @@ namespace Realm {
         register_error_signal_handler(realm_backtrace);
     }
 
+#if defined(REALM_ON_LINUX) || defined(REALM_ON_MACOS) || defined(REALM_ON_FREEBSD)
+    // register SIGTERM handler
+    struct sigaction action;
+    action.sa_handler = realm_sigterm_handler;
+    sigemptyset(&action.sa_mask);
+    action.sa_flags = 0;  // Not using SA_ONSTACK for this handler
+    CHECK_LIBC( sigaction(SIGTERM, &action, 0) );
```
muraj, line 2029: Why use sigaction when we use signal just above?  We should probably be consistent here.

src/realm/logging.cc
```diff
@@ -199,6 +199,11 @@ namespace Realm {
       cfg->stream->flush();
   }
 
+  void flush_all_streams(void)
```
muraj, line 202: Why does this have to be a separate function?

tests/sigterm_test.cc
```diff
@@ -0,0 +1,225 @@
...
+  task_event.wait();
+
+  log_app.info() << "Top level task completed";
+}
+
+int test_sigterm_handling()
+{
+  printf("Starting SIGTERM handling test\n");
+
+  pid_t child_pid = fork();
```
muraj, line 71: Realm doesn't typically support forking processes, and I fail to see why this is needed.  You can just as easily do a regex on the test to verify the output file has the contents you want.

tests/sigterm_test.cc
```diff
@@ -0,0 +1,225 @@
...
+  if (child_pid == 0) {
+    // Child process - run the Realm application
+
+    // Redirect logs to a file for analysis
+    freopen(TestConfig::log_file.c_str(), "w", stderr);
+
+    Runtime rt;
+    rt.init(0, nullptr);
+
+    rt.register_task(TOP_LEVEL_TASK, top_level_task);
```
muraj, line 82: You don't need a "top level task".  Just have the long running task.  Top level tasks are a legion concept that is not something that is required in Realm.

muraj (changes requested):

eddy16112:
We actually already had a PR for reset signals after crashing https://gitlab.com/StanfordLegion/legion/-/merge_requests/960

src/realm/logging.cc
muraj: Okay... I didn't ask an AI for an answer here, I asked the writer of this code.  If you as the writer of this change don't understand the code you wrote, then that calls into question the entire change here.

Regardless, since the core function of this signal handler is to flush the streams (which, imho, is not something that Realm should be doing itself as no other library that I know of does this and there are tools outside of the application to ensure output buffers are flushed), then it makes more sense to just move the signal handler to logging.cc file or just expose the function through the Logger class rather than a standalone function as it is now.  You can see this idea in Logger::configure_from_commandline, set_default_output, etc etc.

tests/CMakeLists.txt
```diff
@@ -447,6 +447,11 @@ target_compile_features(c_external_inst PUBLIC cxx_std_20)
 target_link_libraries(c_external_inst ${TEST_GPU_LIBS})
 
 add_integration_test(c_event_poisoned "${REALM_TEST_DIR}/c/test_event_poisoned.cc")
+add_integration_test(sigterm_test "${REALM_TEST_DIR}/sigterm_test.cc")
+# SIGTERM handling test uses POSIX signals not available on Windows
+if(WIN32)
+  set_property(TEST sigterm_test PROPERTY DISABLED TRUE)
```
muraj, line 453: This doesn't prevent the test from being compiled.  I'm not sure we should even build this test.

src/realm/runtime_impl.cc
```diff
@@ -373,6 +374,39 @@ namespace Realm {
     os.flush();
   }
 
+#if defined(REALM_ON_LINUX) || defined(REALM_ON_MACOS) || defined(REALM_ON_FREEBSD)
+  // Store the original SIGTERM handler so we can call it after flushing
+  // Initialize to default so if no previous handler exists, we restore to default
+  static struct sigaction original_sigterm_handler = { SIG_DFL, {}, 0, nullptr };
+#endif
+
+  static void realm_sigterm_handler(int signal)
```
muraj, line 383: None of this is protected under a `#if defined(REALM_ON_LINUX)...`  Are you sure this compiles under windows?

muraj (changes requested):

PR #313, opened 2025-09-24
inst: start to remove runtime_impl dependence in RegionInstance for unit testing

muraj:


src/realm/inst_impl.cc
```diff
@@ -691,13 +696,10 @@ namespace Realm {
...
 
   /*static*/ Event RegionInstanceImpl::create_instance(
-      const RuntimeImpl *runtime_impl, RegionInstance &inst, Memory memory,
-      InstanceLayoutGeneric *ilg, const ExternalInstanceResource *res,
-      const ProfilingRequestSet &prs, Event wait_on)
+      RegionInstance &inst, MemoryImpl *mem_impl, InstanceLayoutGeneric *ilg,
+      const ExternalInstanceResource *res, const ProfilingRequestSet &prs, Event wait_on)
   {
-    MemoryImpl *m_impl = runtime_impl->get_memory_impl(memory);
-    assert(m_impl != nullptr && "invalid memory handle");
```
eddy16112, line 699: It is better to  keep this assert. Yes, the `m_impl->new_instance` will segfault if `m_impl` is nullptr, but we might see the segfault inside the `new_instance`, not necessary here, which may increase the difficulty to debug.

src/realm/inst_impl.cc
```diff
@@ -507,10 +510,12 @@ namespace Realm {
     //  will be what actually destroys the instance
     log_inst.info() << "instance destroyed: inst=" << *this << " wait_on=" << wait_on;
 
-    MemoryImpl *mem_impl = get_runtime()->get_memory_impl(*this);
+    MemoryImpl *mem_impl = get_runtime()->get_memory_impl(get_location());
     assert(mem_impl != nullptr && "invalid memory handle");
-    RegionInstanceImpl *inst_impl = mem_impl->get_instance(*this);
-    mem_impl->release_storage_deferrable(inst_impl, wait_on);
+    RegionInstanceImpl *inst_impl = mem_impl->get_instance(id);
+    assert(inst_impl != nullptr && "invalid instance handle");
```
eddy16112, line 516: We just replace it with `get_runtime()->get_instance_impl()`

src/realm/mem_impl.h
```diff
@@ -182,6 +182,8 @@ namespace Realm {
 
     void add_module_specific(ModuleSpecificInfo *info);
 
+    RuntimeImpl *get_runtime() const { return runtime_impl; }
```
eddy16112, line 185: 1. name it to `get_runtime_impl`?
2. why do we need to expose the runtime_impl here? I did not see it is used in this PR. IMHO, The RuntimeImpl inside XYZImpl class should only be consumed by the XYZImpl.

apryakhin: agree here - get_runtime_impl is more descriptive and what's the most important consistent with other "impl grabbers".

muraj: @eddy16112 I want to try to remove get_runtime() from within our *Impl classes.  In this case, the idea is that, whenever a RegionInstanceImpl needs a runtime, rather than calling get_runtime(), it will call `mem_impl->get_runtime_impl()` instead.

src/realm/inst_impl.cc
muraj: @apryakhin agreed for an interface call, but this is an internal call with presumably already verified inputs.  If we get a nullptr here, it's most assuredly a bug on our side.  We shouldn't have these checks at every call within the stack, we can use assert to at least document and verify assumptions.

@eddy16112 in either case, we'd still have this function in the callstack and with a core dump of the crash, it would be easy to figure out.

src/realm/mem_impl.h
eddy16112: OK, then why not put a RuntimeImpl inside the RegionInstanceImpl?

PR #318, opened 2025-09-26
Update the event class and add unit tests

ebendler:


tests/unit_tests/hpp_runtime_test.cc
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+#define REALM_NAMESPACE RealmHPP
+
+#include "realm.hpp"
+#include <gtest/gtest.h>
+
+namespace Realm {
+  extern const char *realm_library_version;
```
muraj, line 24: Please compare against REALM_VERSION instead of an internal library version.

muraj (approved):
Small nit on the runtime_test.cc, otherwise LGTM.

PR #323, opened 2025-09-30
Add dynamic loading of ucx and ucc

muraj:


src/realm/ucx/ucc_comm.cc
```diff
@@ -147,50 +226,65 @@ namespace Realm {
...
-        status = ucc_context_progress(context);
-        ucc_check(status);
+      status = UCC_FUNC(ucc_collective_post)(req);
+      if(status != UCC_OK) {
+        goto Done;
       }
 
-      status = ucc_collective_finalize(req);
-      ucc_check(status);
+      status = wait(context, [&req]() { return req->status; });
```
eddy16112, line 280: NIT: it is better to use `ucc_collective_test` even though it is the same as req->status, we do not need to know the implementation of a request.

src/realm/ucx/ucp_internal.cc
```diff
@@ -953,6 +953,8 @@ namespace Realm {
     {
       assert(initialized_boot);
 
+      ucc_comm->UCC_Finalize();
```
eddy16112, line 956: good catch

eddy16112 (approved):

PR #325, opened 2025-10-01
Consolidate loader logic to cross platform class

muraj:
This change adds a Loader class that can be used to easily create a dynamic loader for a library and works across platforms.


apryakhin:
I am ready to approve. Overall that's looking good. I will do a few more passes though.

apryakhin (approved):

PR #327, opened 2025-10-04
clean up assert for machine_implc.cc

eddy16112:


src/realm/compiler_support.h
```diff
@@ -95,11 +95,27 @@
...
 
-// REALM_ASSERT(cond, message) - abort program if 'cond' is not true
+// REALM_ASSERT(cond) - abort program if 'cond' is not true
+#ifdef NDEBUG
 #if defined(__CUDACC__) || defined(__HIPCC__)
-#define REALM_ASSERT(cond, message) assert(cond)
+#define REALM_ASSERT(cond)                                                             \
+  do {                                                                                 \
+    if(!(cond)) {                                                                      \
+      abort();                                                                         \
```
muraj: IIRC there is no abort in cuda, so this probably won't compile...

src/realm/compiler_support.h
```diff
@@ -95,11 +95,27 @@
...
+    }                                                                                  \
+  } while(0)
 #else
-#define REALM_ASSERT(cond, message) assert((cond) && (message))
+#define REALM_ASSERT(cond)                                                             \
+  do {                                                                                 \
+    if(!(cond)) {                                                                      \
+      fprintf(stderr, "Assertion failed: (%s), at %s:%d", #cond, __FILE__, __LINE__);  \
```
muraj: We shouldn't be using fprintf / fflush / stderr here.  We should probably use the runtime logger instead.

eddy16112: Unfortunately, we can not include logger.h here.

muraj: Why not?

src/realm/compiler_support.h
eddy16112: Good catch, I replaced it with `__trap()`

src/realm/compiler_support.h
eddy16112: OK, I fixed it

Issue #331, opened 2025-10-09
[BUG] Finalization segfault if using Open MPI 5


eddy16112:
**Describe the bug**
We have seen a segfault during runtime finalization when using UCX+Open MPi 5

**To Reproduce**
1. Create a conda env with Open MPI 5.0.8
```
conda create -n ompi5
conda activate ompi5
conda install conda-forge::openmpi
```
2. Compile realm with UCX
```
cmake ../ -DREALM_ENABLE_UCX=ON -DREALM_ENABLE_CUDA=OFF -DREALM_ENABLE_HIP=OFF -DREALM_ENABLE_OPENMP=OFF -DREALM_ENABLE_PYTHON=ON -DREALM_ENABLE_HDF5=OFF -DREALM_BUILD_TESTS=ON
```

3. Run any of the realm program, I picked memspeed and remove everything except runtime init and shutdown.
```
memspeed
```

**Expected behavior**
Here is the backtrace
```
Thread 6 (Thread 0x7fffe0fddc00 (LWP 1607363) "memspeed" (Exiting)):
#0  0x00007ffff73af5b0 in ?? ()
#1  0x00007ffff7894be1 in advise_stack_range (guardsize=<optimized out>, pd=140736968121344, size=<optimized out>, mem=0x7fffe0edb000) at ./nptl/allocatestack.c:195
#2  start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:551
#3  0x00007ffff79268c0 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81

Thread 5 (Thread 0x7ffff01a2c00 (LWP 1607362) "memspeed" (Exiting)):
#0  0x00007ffff73af5b0 in ?? ()
#1  0x00007ffff7894be1 in advise_stack_range (guardsize=<optimized out>, pd=140737221635072, size=<optimized out>, mem=0x7ffff00a0000) at ./nptl/allocatestack.c:195
#2  start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:551
#3  0x00007ffff79268c0 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81

Thread 3 (Thread 0x7ffff13ff640 (LWP 1607357) "cuda0000380000f"):
#0  0x00007ffff7918c3f in __GI___poll (fds=0x555555f15e50, nfds=3, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
[...]

PR #333, opened 2025-10-10
Fix the segfault when using Open MPI 5 to bootstrap UCX

eddy16112:
Please see https://github.com/StanfordLegion/realm/issues/331 for the bug.

### DIFF
--- a/src/realm/ucx/bootstrap/bootstrap_mpi.c
+++ b/src/realm/ucx/bootstrap/bootstrap_mpi.c
@@ -235,11 +235,12 @@ int realm_ucp_bootstrap_plugin_init(void *mpi_comm, bootstrap_handle_t *handle)
     fprintf(stderr, "dladdr returned NULL dli_fname\n");
     return -1;
   }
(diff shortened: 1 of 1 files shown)


src/realm/ucx/bootstrap/bootstrap_mpi.c
```diff
@@ -221,6 +224,23 @@ static int bootstrap_mpi_finalize(bootstrap_handle_t *handle)
...
+  }
+  if(info.dli_fname == NULL) {
+    fprintf(stderr, "dladdr returned NULL dli_fname\n");
+    return -1;
+  }
+  if(dlopen(info.dli_fname, RTLD_NOW | RTLD_NODELETE) == NULL) {
+    fprintf(stderr, "dlopen failed for %s: %s\n", info.dli_fname, dlerror());
+    return -1;
+  }
+  dlclose(handle);
```
marcinz: Is the intention here to actually close `handle`?

eddy16112: My understanding is the dlclose is just to decrease the reference count of the handle.

muraj (approved):

PR #336, opened 2025-10-22
HPP: refresh Machine class and MachineQueryIterator class

ebendler:


muraj (changes requested):

src/realm.hpp
```diff
@@ -1830,17 +1844,20 @@ namespace REALM_NAMESPACE {
...
     MachineQueryIterator<QT, RT> &operator++(/*prefix*/)
     {
-      throw std::logic_error("Not implemented");
+      result = query_ptr.next(result);
+      return *this;
     }
 
     MachineQueryIterator<QT, RT> operator++(int /*postfix*/)
     {
-      throw std::logic_error("Not implemented");
```
muraj: I did not see any unit tests for this added behavior.  In fact, the entire class is not even compiled, as evidenced by the following codecov link:

https://app.codecov.io/gh/StanfordLegion/realm/commit/016fabc7bd40f41bd6672bdd0a9b5a699a25d62f/blob/src/realm.hpp?dropdown=coverage#L1790

If we're going to be adding new functionality to this class in this PR, then it needs to be unit tested.

ebendler: This class gets used by the proc/mem query classes. I'll just move it to those PRs.

tests/unit_tests/hpp_machine_tests.cc
```diff
@@ -0,0 +1,524 @@
...
+  EXPECT_EQ(localProcessors.size(), config.getLocalProcessorCount(0));
+
+  // Verify all are in local address space
+  for(const auto &proc : localProcessors) {
+    EXPECT_TRUE(proc.exists());
+    EXPECT_EQ(proc.address_space(), 0);
+  }
+}
+
+TEST_P(HPPMachineTest, QueryLocalProcessorsByCPUKind)
```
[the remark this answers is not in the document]
ebendler: I've updated the test to do more with instantiated topologies

Issue #339, opened 2025-10-29
Realm registers whole `fbmem` and fails when it's too large


marcinz:
After fixes in #337, I was able to run Legate on Perlmutter using the `mpich` module, but the following problem popped up:

```
*** WARNING (proc 1): Unexpected error 12 (Cannot allocate memory) from fi_mr_regattr() when binding segment [0x602000000, 0xecc000000) to EP 2
*** WARNING (proc 1): GASNet gex_EP_BindSegment returning an error code: GASNET_ERR_RESOURCE (Problem with requested resource)
  from function gasnetc_ep_bindsegment_hook(i_ep, i_segment, flags)
  at /pscratch/sd/m/mzalewsk/conda_envs/legate-25.10.rc7/gex-wrapper/src/build/embed-gasnet/source/gasnet_mmap.c:2144
  reason: GASNet encountered an error: GASNET_ERR_RESOURCE(10002)
failed to bind segment*** Caught a fatal signal (proc 1): SIGABRT(6)
*** NOTICE (proc 1): Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace.
*** NOTICE (proc 1): We recommend linking the debug version of GASNet to assist you in resolving this application issue.
[1 - 7fb069131740]    0.000000 {2}{gexbind}: segment bind?: base=0x7f747ffff100 size=20971520000 mtype=1 extra=0
[1 - 7fb069131740]    0.000000 {2}{gexbind}: segment bound: base=0x7f747ffff100 size=20971520000 mtype=1 extra=0 ep_index=1
[1 - 7fb069131740]    0.000000 {2}{gexbind}: segment bind?: base=0x602000000 size=37748736000 mtype=2 extra=94504547508144
*** WARNING (proc 0): Unexpected error 12 (Cannot allocate memory) from fi_mr_regattr() when binding segment [0x602000000, 0xecc000000) to EP 2
*** WARNING (proc 0): GASNet gex_EP_BindSegment returning an error code: GASNET_ERR_RESOURCE (Problem with requested resource)
  from function gasnetc_ep_bindsegment_hook(i_ep, i_segment, flags)
  at /pscratch/sd/m/mzalewsk/conda_envs/legate-25.10.rc7/gex-wrapper/src/build/embed-gasnet/source/gasnet_mmap.c:2144
  reason: GASNet encountered an error: GASNET_ERR_RESOURCE(10002)
[...]

PR #346, opened 2025-10-31
Remove abort path from a failed rdma registration in gasnet

muraj:


muraj:
@marcinz this should remove the crash for issue #339

src/realm/gasnetex/gasnetex_internal.cc
```diff
@@ -3375,14 +3375,17 @@ namespace Realm {
       }
     }
 
+    if(gex_wrapper_handle.gex_ep_bind_segment(ep, segment, 0 /*flags*/) != 0) {
+      // TODO: destroy the segment and ep we created?
+      return false;
```
eddy16112: Let's add a warning message here.

eddy16112: > Enabling gasnet warnings when debugging this is enough for now I think.

We do not have warnings if bind segment is failed.

PR #349, opened 2025-11-05
Add PMIX

apryakhin:


eddy16112 (commented):
@muraj Do we have ucx test in our CI? If so, we probably need to add the PMIx to the ucx test as well.

src/realm/ucx/bootstrap/CMakeLists.txt
```diff
@@ -72,6 +72,17 @@ if(UCX_BOOTSTRAP_ENABLE_MPI)
     )
   endif()
 endif()
+
+if(UCX_BOOTSTRAP_ENABLE_PMIX)
+  find_package(PMIX)
```
eddy16112: there is no FindPMIX.cmake, so I am getting the following error
```
CMake Warning at src/realm/ucx/bootstrap/CMakeLists.txt:77 (find_package):
  By not providing "FindPMIX.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "PMIX", but
  CMake did not find one.

  Could not find a package configuration file provided by "PMIX" with any of
  the following names:

    PMIXConfig.cmake
    pmix-config.cmake

  Add the installation prefix of "PMIX" to CMAKE_PREFIX_PATH or set
  "PMIX_DIR" to a directory containing one of the above files.  If "PMIX"
  provides a separate development package or SDK, be sure it has been
  installed.
```
We also need to make the PMIX as required if PMIX is enabled.

muraj: No, please don't make pmix required if enabled.  The way we set up dependencies here is, if enabled, the build system will look for the dependency and use it if it finds it.  Otherwise it'll disable the feature and not use it.  So please, no required.

If you need FindPMIX.cmake, I think pkg-config should find it, so copy what we do with gasnet and hwloc, etc.

src/realm/ucx/bootstrap/bootstrap.cc
```diff
@@ -21,6 +21,8 @@
 #include "realm/ucx/bootstrap/bootstrap_internal.h"
 #include "realm/ucx/bootstrap/bootstrap_loader.h"
 
+#include <iostream>
```
elliottslaughter, line 24: Why are we including `iostream`? Doesn't appear to be used.

elliottslaughter:
@artempriakhin What's the status on this?

src/realm/ucx/bootstrap/CMakeLists.txt
elliottslaughter: We should update this to follow the style in #434. That is:

 * `find_package(PMIX QUIET)` to discover if we have PMIx
 * use `PMIX_FOUND` as the default for `option(UCX_BOOTSTRAP_ENABLE_PMIX ...)`
 * under `if(UCX_BOOTSTRAP_ENABLE_PMIX)` now `find_package(PMIX REQUIRED)` so that if the user explicitly requested, we'll hit an error if it's not found

elliottslaughter: Also, looks like the `option()` call is currently missing so we won't get a cache variable right now.

PR #365, opened 2025-11-20
fixing CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES error

ipdemes:
this PR fixes 701 error discovered in cuPyNumeric tests when running multi-GPU:
The CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES (error 701) was caused by parameter passing incompatibility when using CU_LAUNCH_PARAM_BUFFER_POINTER with small (1-byte) reduction operator structs. The packed parameter buffer didn't meet CUDA's alignment expectations, causing the driver to reject kernel launches even with minimal resources (I debugged with 1 thread, 15 registers and test still failed) where occupancy APIs indicated success.
The fix involved two changes: switching from buffer pointer to params array method for kernel launches (which passes pointers to each parameter instead of a packed buffer), and implementing per-GPU kernel caching with proper context management for multi-GPU systems. These changes ensure correct parameter alignment and that each GPU uses context-specific kernel instances, resolving the launch failures on multi-GPU configurations.

### DIFF
--- a/src/realm/cuda/cuda_internal.cc
+++ b/src/realm/cuda/cuda_internal.cc
@@ -2248,23 +2248,23 @@ namespace Realm {
                                             : redop->cuda_fold_nonexcl_fn)
                  : (redop_info.is_exclusive ? redop->cuda_apply_excl_fn
                                             : redop->cuda_apply_nonexcl_fn));
-        
+
         if(redop->cudaGetFuncBySymbol_fn != 0) {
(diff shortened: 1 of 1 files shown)


lightsighter:
> How should I do code formatting?

Run this command here: https://github.com/StanfordLegion/realm/actions/runs/19553245435/job/55989723700?pr=365#step:6:20 until all the diffs go away.

Issue #370, opened 2025-12-03
Build: undefined assert on macOS


elliottslaughter:
Since #369 we're hitting this error on macOS:

```
/Users/gitlabrunner/builds/WfUNgfK_6/0/StanfordLegion/legion/realm/install/include/realm/point.inl:99:5: error: use of undeclared identifier 'assert'
    assert(index < N);
    ^
```

Job: https://gitlab.com/StanfordLegion/legion/-/jobs/12314317975

Blocks Realm update to Legion: https://gitlab.com/StanfordLegion/legion/-/merge_requests/1960

eddy16112:
Here is a fix https://github.com/StanfordLegion/realm/pull/371

elliottslaughter:
Never mind, those are Legion code. Please disregard.

PR #371, opened 2025-12-03
Fix the assert in bytearray, point and utils

eddy16112:
This PR fixes the https://github.com/StanfordLegion/realm/issues/370.
1. Instead of include the `assert.h`, I just replace all the `assert` with `REALM_ASSERT`. 
2. We can not include `logging.h` before `REALM_ASSERT`, because `logging.h` includes `utils.h`, which has calls to `REALM_ASSERT`, so instead of calling logger inside `REALM_ASSERT`, I wrapper the logger into a function.


apryakhin (approved):

src/realm/sparsity.inl
```diff
@@ -20,7 +20,7 @@
 // nop, but helps IDEs
 #include "realm/sparsity.h"
 
-#include "realm/assert.h"
+#include "realm/realm_assert.h"
```
apryakhin: nit: Why did you rename it from "assert.h" to "realm_assert.h" ? Unless that's conflated with the og assert head somehow, I think we are fine with just a "realm/assert.h"...naming-wise it's already our custom assert under "realm" folder, so it's self-obvious that it's a realm custom assert.

apryakhin (approved):

Issue #373, opened 2025-12-09
Nix build reports that realm.pc is broken


elliottslaughter:
A fairly straightforward Nix derivation for Realm results in the following error:

```
Broken paths found in a .pc file! /nix/store/qmvh8ha0nrl9cs72k7zrj3czw96hhcjh-realm-2025-12-03/share/pkgconfig/realm.pc
The following lines have issues (specifically '//' in paths).
18:libdir="${prefix}//nix/store/qmvh8ha0nrl9cs72k7zrj3czw96hhcjh-realm-2025-12-03/lib"
19:includedir="${prefix}//nix/store/qmvh8ha0nrl9cs72k7zrj3czw96hhcjh-realm-2025-12-03/include"
It is very likely that paths are being joined improperly.
ex: "${prefix}/@CMAKE_INSTALL_LIBDIR@" should be "@CMAKE_INSTALL_FULL_LIBDIR@"
Please see https://github.com/NixOS/nixpkgs/issues/144170 for more details.
```

Note: I think the suggested fix is **wrong** and should NOT be applied to Realm. If you follow the link to the issue you'll see that if you do things this way you'll lose relocatable builds for everything non-Nix related.

Looking at the generated `realm.pc`, I see the following:

```
prefix=/nix/store/qmvh8ha0nrl9cs72k7zrj3czw96hhcjh-realm-2025-12-03
exec_prefix="${prefix}"
libdir="${prefix}//nix/store/qmvh8ha0nrl9cs72k7zrj3czw96hhcjh-realm-2025-12-03/lib"
includedir="${prefix}//nix/store/qmvh8ha0nrl9cs72k7zrj3czw96hhcjh-realm-2025-12-03/include"
datarootdir="${prefix}/share"
datadir="${prefix}/share"

Name: Realm
Description: A distributed, event-based tasking library
Version: 25.6.1-rc.4-21-g9de4b38680
URL: https://legion.stanford.edu/realm
Requires.private: threads cudatoolkit python3
Cflags: -I${includedir} 
Libs: -L${libdir} -lrealm
```

What I find interesting is that `data*dir` are different than `libdir` and `includedir`. This makes me think that the following logic is responsible:

https://github.com/StanfordLegion/realm/blob/9de4b38680995b726bb8a43264a2d2a9870e1f1c/cmake/RealmPackaging.cmake#L122

However, what I don't know is whether removing this logic will impact the standard case. I don't want to do something that will break Realm for non-Nix users.

[...]

PR #375, opened 2025-12-09
Fix path normalization for pkgconfig to account for absolute paths

elliottslaughter:
Fixes https://github.com/StanfordLegion/realm/issues/373

pkgconfig paths are tricky because (a) the end user may relocate the installation, and (b) when invoking CMake, the user may potentially specify absolute paths for both `CMAKE_INSTALL_PREFIX` and `CMAKE_INSTALL_LIBDIR`, leading to situations where one is not contained in the other.

This patch does a couple of things:

 * If libdir (and friends) are relative, install as `${prefix}/libdir`
 * If libdir is absolute but contained within the installation prefix, make it relative and then do the above
 * If libdir is absolute and NOT contained within the installation prefix, just keep it as it. The installation will not be relocatable but I'm not sure what else can be done in this scenario

I think this covers all the possible cases.

Tested by hand with several different configuration options, such as setting libdir to something entirely unrelated (e.g., `/a/b/c`).


elliottslaughter:
Example running on my local Mac:

```console
$ cmake -DCMAKE_INSTALL_PREFIX=$PWD/../install -DCMAKE_INSTALL_LIBDIR=$PWD/../install/lib ..
$ make install -j4
$ cat ../install/share/pkgconfig/realm.pc
...

prefix=/Users/elliott/Programming/Legion/realm/install
exec_prefix="${prefix}"
libdir="${prefix}/lib"
includedir="${prefix}/include"
datarootdir="${prefix}/share"
datadir="${prefix}/share"

Name: Realm
Description: A distributed, event-based tasking library
Version: 25.6.1-rc.4+bfbd7b76
URL: https://legion.stanford.edu/realm
Requires.private: threads python3
Cflags: -I${includedir} 
Libs: -L${libdir} -lrealm 
```

elliottslaughter:
@muraj Do you expect everyone to sign their commits or can this be merged as-is?

PR #377, opened 2025-12-12
Add MPI based vtable example

sbahirnv:
Adds a MPI-based NetworkVtable example that demonstrates how to implement the bootstrap interface. 

The example builds and runs with `mpirun -n 2` successfully.

### DIFF
--- a/examples/vtable/realm_bootstrap.cc
+++ b/examples/vtable/realm_bootstrap.cc
@@ -13,19 +13,25 @@ static constexpr const char *REALM_KEY_RANK = "realm_rank";
 static constexpr const char *REALM_KEY_RANKS = "realm_ranks";
 static constexpr const char *REALM_KEY_GROUP = "realm_group";
 
-static std::map<std::string, std::vector<uint8_t>> local_kv_store;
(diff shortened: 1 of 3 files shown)


examples/vtable/realm_bootstrap.cc
```diff
@@ -0,0 +1,171 @@
...
+  *value_size = 0;
+  return true;
+}
+
+static bool app_bar(const void *vtable_data, size_t vtable_data_size)
+{
+  // std::cout << "app_bar called" << std::endl;
+  
+  if(MPI_Barrier(MPI_COMM_WORLD) != MPI_SUCCESS)
+    return false;
```
lightsighter: I don't think you actually need this, the blocking all-gather call farther down also functions as a barrier across the processes because it does not return until it receives something from every other process which means that everyone has arrived at it.

sbahirnv (commented):
Addressed review comments, please take a look.

apryakhin:
@sbahirnv are you planning on getting this merged?

lightsighter:
> @sbahirnv are you planning on getting this merged?

You can merge it into my branch whenever you're ready. We can clean up things more in my branch before anything hits main.

PR #382, opened 2025-12-18
Remove cflags from gasnet find module, as they override many of the f…

muraj:
…lags set by cmake for no real reason


elliottslaughter:
Legion CI is happy with it, and it does seem to meaningfully reduce build times: https://gitlab.com/StanfordLegion/legion/-/pipelines/2222815558/

elliottslaughter (approved):

PR #387, opened 2026-01-13
add missing symbol visibility for realm_kokkos

rbberger:
Linking to `librealm.so` failed because `realm_kokkos` didn't export `kokkos_initialize` and `kokkos_finalize`. This PR adds the missing `REALM_PUBLIC_API` annotations.


rbberger:
@muraj The symbols from `librealm.so` are needed by `librealm_kokkos.so`. The symbols from `librealm_kokkos.so` are needed by `librealm.so`. Clients linking to `librealm.so` without these exports are unhappy.

rbberger:
@eddy16112 pushed and signed.

elliottslaughter:
Are we good to merge this then?

PR #390, opened 2026-01-20
Bootstrap vtable pmix example

sbahirnv:
Adds a PMIx-based KeyValueStoreVtable example that demonstrates how to implement the bootstrap interface.

The example builds and runs with `prterun -np 2`. 
Note: there is a known PMIx caching issue that will cause bootstrap failures (https://github.com/openpmix/openpmix/tree/v5.0). Fix available in PMIx 5.0.10.

### DIFF
--- a/examples/vtable/pmix/realm_bootstrap.cc
+++ b/examples/vtable/pmix/realm_bootstrap.cc
@@ -23,10 +23,7 @@
 
 namespace App {
 
-  // Realm KeyValueStoreVtable required keys
-  static constexpr const char *REALM_KEY_RANK = "realm_rank";
-  static constexpr const char *REALM_KEY_RANKS = "realm_ranks";
-  static constexpr const char *REALM_KEY_GROUP = "realm_group";
+  using KVVtable = Realm::Runtime::KeyValueStoreVtable;
 
   struct VtableContext {
     // pending_sync is an optimization to avoid redundant PMIx_Commit() calls
@@ -67,7 +64,7 @@ namespace App {
 
     std::string k(static_cast<const char *>(key), key_size);
 
-    if(k == REALM_KEY_RANK) {
+    if(k == KVVtable::rank_key) {
       if(*value_size < sizeof(uint32_t)) {
         *value_size = 0;
         return false;
@@ -77,7 +74,7 @@ namespace App {
       *value_size = sizeof(v);
       return true;
     }
-    if(k == REALM_KEY_RANKS) {
+    if(k == KVVtable::ranks_key) {
       if(*value_size < sizeof(uint32_t)) {
         *value_size = 0;
         return false;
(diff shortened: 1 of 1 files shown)


src/realm/ucx/bootstrap/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ if(NOT Realm_VERSION_MAJOR)
 endif()
 
 option(UCX_BOOTSTRAP_ENABLE_MPI "Compile with MPI support" OFF)
+option(UCX_BOOTSTRAP_ENABLE_PMIX "Compile with PMIx support" OFF)
```
lightsighter: I don't think this should be in the UCX module part of cmake. I feel like this is a property of the test. If you build that particular test then you need PMIx support, but the UCX module itself doesn't inherently need it.

examples/vtable/pmix/realm_bootstrap.cc
```diff
@@ -0,0 +1,241 @@
...
+    PMIX_VALUE_RELEASE(val);
+
+    ctx->pmix_rank = static_cast<uint32_t>(ctx->myproc.rank);
+
+    Realm::Runtime::KeyValueStoreVtable vtable;
+    vtable.vtable_data = new VtableContext *(ctx);
+    vtable.vtable_data_size = sizeof(VtableContext *);
+    vtable.put = app_put;
+    vtable.get = app_get;
+    vtable.bar = app_bar;
```
apryakhin: Where are we using it?

examples/vtable/pmix/realm_bootstrap.cc
```diff
@@ -0,0 +1,241 @@
...
+
+    std::string k(static_cast<const char *>(key), key_size);
+    std::vector<uint8_t> v(static_cast<const uint8_t *>(value),
+                           static_cast<const uint8_t *>(value) + value_size);
+    ctx->local_kv_store[k] = v;
+    ctx->pending_sync = true;
+    return true;
+  }
+
+  static bool app_get(const void *key, size_t key_size, void *value, size_t *value_size,
```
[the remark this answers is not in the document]
sbahirnv: > Where is PMIX_get here?

If you mean in the implementation of the `app_get()` ,  its down there, towards the end of function.

> You have either explained it somewhere I wan't part of or this PR is really missing some context. 

All the discussion happened either on the PR, slack or realm standup only and AFAIR you were present in all of them :) 

> I don't even understand what's all this implemented here?

It  an implementation of `app_get()` API from interface which is well documented here: https://github.com/StanfordLegion/realm/blob/11f29ee3a43a1f732477ce51bd76b255721ad5fb/src/realm/runtime.h#L100

Exactly which specific part did you not understand?

src/realm/ucx/bootstrap/CMakeLists.txt
sbahirnv: Agreed. Moved PMIx dependency to the example only.

sbahirnv (commented):
Addressed the comments, please take a relook
^^ @lightsighter , @apryakhin , @eddy16112

PR #391, opened 2026-01-21
Prototype of realm subgraph compilation approach

rohany:
This PR contains the Realm components of the compilation approach described in https://rohany.github.io/publications/actor-task-duality.pdf. The goal of this PR is not to merge the code directly but to start a conversation about the overall approach, and what major steps we feel like the approach needs in order to be something that we can commit to in the future. Given that, the code itself should likely go through some major refactoring before it is merged, so this PR is here to describe the overall approach, rather than being committed to a particular implementation.


src/realm/tasks.cc
```diff
@@ -1129,6 +1153,252 @@ namespace Realm {
           update_worker_count(0, +1);
         }
 
+  // If we're not working on a subgraph right now, see if there's
+  // anything to pop on the pending subgraph queue.
+  if (current_subgraph == nullptr) {
```
apryakhin: I don't think leaking the subgraph details into the scheduler is a good idea. This now becomes very fragile with potential functionality changes to the subgraph risking to break this. We need to abstract this away into some form of `SubgraphExecutor` at least.

PR #393, opened 2026-01-22
Key Value Store Vtable Strings

lightsighter:
Provide static constexpr versions of built-in key strings for clients to use that are guaranteed to match what Realm uses internally.

### DIFF
--- a/src/realm/runtime_impl.cc
+++ b/src/realm/runtime_impl.cc
@@ -1352,17 +1352,17 @@ namespace Realm {
 
   std::optional<uint64_t> RuntimeImpl::key_value_store_local_group(void) const
   {
-    return key_value_store_get_int(KeyValueStoreVtable::group_key);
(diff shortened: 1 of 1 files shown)


src/realm/runtime.h
```diff
@@ -157,6 +157,9 @@ namespace Realm {
        * across groups. Realm will generate a unique address space for
        * each process as part of the bootstrap.
        */
+      inline static constexpr std::string_view group_key = "realm_group";
+      inline static constexpr std::string_view rank_key = "realm_rank";
+      inline static constexpr std::string_view ranks_key = "realm_ranks";
```
sbahirnv: [nit] Given that vtable function pointers that are C compatible, these string_view constants won't be usable from pure C in case there is any intended C user.

lightsighter: This is the C++ API, but you can always call `string_view::data()` to get a raw pointer to the underlying string and `string_view::size()` to get the size of the string.

PR #394, opened 2026-02-02
Initial unit tests and test harness for subgraphs

rohany:
Starting to write a set of unit tests (with more tests in the TODOs for follow up PRs) to assist with the pending compiled subgraph implementation.


rohany:
@artempriakhin can you take a look at this PR? I'm getting somewhat close to having something to test and having unit tests already in the repo that work / are blessed would be good.

PR #397, opened 2026-02-08
Best Affinity Fixes

lightsighter:
This branch fixes some bugs that slipped through the original best affinity pull request. They were caught when running the Legion CI (which suggests that there are tests and configurations of Realm that Legion CI is still covering that the Realm CI is not covering).


elliottslaughter:
Would it be possible to increase the coverage of the native Realm tests so that they catch the case we're fixing? I'm glad we have coverage via Legion, but I'm concerned that this will break easily if we're not actually testing the behavior on the Realm side.

PR #402, opened 2026-02-25
Fix the redop path when p2p is disable

eddy16112:
When p2p between GPU 0 and GPU 1 is disabled, the GPU 1 is not able to access the ZC Memory pinned to GPU 0, so we need to disable this path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -522,7 +522,35 @@ if(REALM_ENABLE_KOKKOS)
   # the user to install it themselves.
   find_package(Kokkos REQUIRED COMPONENTS separable_compilation)
   enable_language(${Kokkos_COMPILE_LANGUAGE})
-  list(APPEND REALM_LIBRARIES realm_kokkos)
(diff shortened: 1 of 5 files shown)


src/realm/cuda/cuda_module.cc
```diff
@@ -4778,32 +4778,41 @@ namespace Realm {
...
-        desc.push_back({});
         desc.back().flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
         desc.back().location.type = CU_MEM_LOCATION_TYPE_HOST_NUMA;
         desc.back().location.id = 0;
 #else
         res = CUDA_ERROR_NOT_SUPPORTED;
         goto Done;
 #endif // CUDA_VERSION >= 12000
+      } else {
+        if(!peer_enabled) {
```
muraj: The mapping to the local device should always happen no matter what.  Are you sure that gpu->info->peers has the local device in it?

PR #406, opened 2026-03-04
Background Copy Planning

lightsighter:
This pull request moves all copy planning into a background work item so that all copy launches appear as close to O(1) as possible and do not consume any client execution time for doing things like path planning. The only analysis that is done currently in the client's thread is the analysis to determine if there are any additional events that need to be folded into the precondition for the copy before it is deferred or started.

### DIFF
--- a/src/realm/bgwork.cc
+++ b/src/realm/bgwork.cc
@@ -389,22 +389,24 @@ namespace Realm {
                        << " slot=" << index;
 #ifdef DEBUG_REALM
     State old_state = state.exchange(STATE_ACTIVE);
-    if(old_state != STATE_IDLE) {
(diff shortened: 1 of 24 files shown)


eddy16112:
What is the motivation of this PR (what performance issue did you see)? My concern is that is it too fine grain if we move the copy analysis into a bgwork item? We did not measure the METG of the bgwork before, if a work is smaller than the METG, I am worried that the overhead will negate the performance benefits

apryakhin (approved):

muraj (approved):

elliottslaughter:
This made it through a fuzzer run (in combination with #428). Looks good from my side. Merge when you're ready.

elliottslaughter (approved):
Passed fuzzer run.

PR #410, opened 2026-03-10
Fix Active Message Payload Fragmentation

lightsighter:
Add automatic payload chunking to the ActiveMessage class so that any active message whose payload exceeds the network backend's hard limit is transparently fragmented and reassembled, eliminating crashes like the assert(size <= ib_seg_size) failure in the UCX backend.                                                                                                                            
                                                                                                                                        
  Changes                                                                                                                             

  New NetworkModule::max_payload_size() interface
  - Added a max_payload_size(size_t header_size) pure virtual method to NetworkModule and a corresponding free function in the Network namespace. This returns the strict upper bound on payload size for a single active message on the eager (no RDMA) path, independent of congestion or buffer registration.
  - GASNet-EX: returns the medium message limit derived from AM_LUBRequestMedium and cfg_outbuf_size.
  - GASNet-1: returns gasnet_AMMaxMedium().
  - MPI: returns AM_BUF_SIZE - header_size.
  - UCX: returns SIZE_MAX since UCX handles fragmentation internally via automatic eager/rendezvous protocol selection.
  - Loopback: returns SIZE_MAX.

  Automatic chunking in ActiveMessage
  - ActiveMessage::init(NodeID, size_t) now checks the requested payload size against Network::max_payload_size(). If the payload exceeds the limit, the message enters a "chunked mode" that buffers payload locally and splits it into fragments at commit() time.
  - Each fragment is sent as an ActiveMessage<WrappedWithFragInfo<T>> with FragmentInfo metadata (chunk ID, total chunks, message ID). The existing IncomingMessageManager / FragmentedMessage infrastructure reassembles them transparently before invoking the handler.
  - For the common case where the payload fits within the network limit, the code path is identical to before — zero overhead. On UCX (which returns SIZE_MAX), the chunking path is never entered.

  Automatic dual handler registration
  - ActiveMessageHandlerReg<T> now automatically registers a WrappedWithFragInfo<T> handler alongside the plain T handler. This ensures every message type can be received as a fragmented message without requiring explicit opt-in. The WrappedWithFragInfo handler uses the existing wrap_handler_unwrap mechanism to strip the FragmentInfo and dispatch to the original handler.

  UCX pbuf_get assertion removal
  - Removed assert(size <= ib_seg_size) from UCPInternal::pbuf_get(). This was a Realm-side guard, not a UCX requirement. UCX's UCP layer handles message fragmentation internally — the send path automatically selects eager vs rendezvous based on message size, and the receive path already fully supports rendezvous via UCP_AM_RECV_ATTR_FLAG_RNDV.

  Removal of ActiveMessageAuto
  - Deleted the ActiveMessageAuto class, DefaultActiveMessageBuilder type alias, and AutoMessageRegistrar struct, since their functionality is now subsumed by the base ActiveMessage class.
  - Converted the sole usage site in barrier_impl.cc (BARRIER_ENABLE_BROADCAST path) from ActiveMessageAuto<BarrierTriggerMessage> to plain ActiveMessage<BarrierTriggerMessage>.
  - Removed the AutoMessageRegistrar<BarrierTriggerMessage> instance (automatic dual-registration handles this now).
  - Deleted tests/unit_tests/auto_actmsg_test.cc and removed it from CMakeLists.txt.

@SeyedMir @apryakhin


lightsighter:
Running Legion CI: https://gitlab.com/StanfordLegion/legion/-/pipelines/2377414546

src/realm/network.h
```diff
@@ -137,6 +137,13 @@ namespace Realm {
                                    size_t line_stride,
                                    const RemoteAddress &dest_payload_addr,
                                    bool with_congestion, size_t header_size);
+
+    // returns the strict upper bound on payload size for a single active
+    //  message on the simplest (eager, no RDMA) path - this is independent
+    //  of congestion, source/dest registration, etc.
+    // network backends that handle fragmentation internally (e.g. UCX) may
+    //  return SIZE_MAX to indicate no practical limit
+    size_t max_payload_size(size_t header_size);
```
SeyedMir: I think this function needs at least another argument to specify whether the user will provide the payload buffer or not. Maybe pass a src buffer pointer (similar to recommended_max_payload) and set it to NULL if the query is for a case that the network module should provide the payload buffer.

SeyedMir: Looks reasonable to me; config.pbuf_max_size should be what is returned for UCX module when src_payload_addr is nullptr. Note that the current default value for pbuf_max_size is 8KB.
https://github.com/StanfordLegion/realm/blob/ea5277741575d81501cb0c40a3c4f78926a3cbd8/src/realm/ucx/ucp_internal.h#L122

lightsighter: > Looks reasonable to me; config.pbuf_max_size should be what is returned for UCX module when src_payload_addr is nullptr. Note that the current default value for pbuf_max_size is 8KB.

I had Claude make this fix. See if it looks good to you.

> It is worth documenting for the user of the new API that using max_payload_size may lead to sub-optimal perf (and that's why we have recommended_max_payload).

I added documentation to that effect. See if you are happy with it.

src/realm/network.inl
```diff
@@ -219,6 +219,15 @@ namespace Realm {
             dest_payload_addr, with_congestion, header_size);
     }
 
+    inline size_t max_payload_size(size_t header_size, const void *src_payload_addr)
+    {
+#ifdef REALM_USE_MULTIPLE_NETWORKS
```
apryakhin: This needs at least a TODO on what to do here next

lightsighter: Claude is just mirroring what exists in related functions. You can see them in the `main` branch here:

https://github.com/StanfordLegion/realm/blob/main/src/realm/network.inl#L144-L220

What do you think should go in there for all those different functions?

src/realm/activemsg.h
```diff
@@ -170,6 +173,21 @@ namespace Realm {
     T *header;
     Realm::Serialization::FixedBufferSerializer fbs;
     uint64_t inline_capacity[INLINE_STORAGE / sizeof(uint64_t)];
+
+    // chunked-mode state: only used when the requested payload exceeds the
+    //  network's hard limit for a single message
+    size_t network_max_payload_; // 0 = normal (non-chunked) mode
+    NodeID chunk_target_;
+    std::vector<char> chunk_buffer_;
```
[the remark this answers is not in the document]
lightsighter: Claude and I went through and removed all the explicit chunking logic from active messages that used to do that on the calling side. The unnecessary chunk_buffer is now gone and all chunking logic is now encapsulated exclusively inside the `ActiveMessage` class.

src/realm/network.inl
eddy16112: I would recommend to add an assert(0) here

src/realm/activemsg.inl
```diff
@@ -181,6 +221,7 @@ namespace Realm {
                                               size_t _line_stride)
   {
     assert(impl == 0);
+    network_max_payload_ = 0;
```
eddy16112: Do we need `init_chunked_data` here?  If it is not supported, let's add a TODO here.

src/realm/activemsg.inl
```diff
@@ -94,10 +112,13 @@ namespace Realm {
                                               size_t _max_payload_size /*= 0*/)
   {
     assert(impl == 0);
+    // NodeSet chunking not yet supported - assert payload fits
+    // TODO: implement chunked mode for multi-target messages
```
eddy16112: Is there anything special we need for multi-target messages?

src/realm/network.inl
lightsighter: This code is the same as what's in the `main` branch. Do you want me to do that for all those functions? It's not even clear to me how this code is correct in the `main` branch if the macro is defined because some of these functions need to return values and if the macro is defined they don't return anything.

src/realm/activemsg.inl
lightsighter: The 2D payload active message constructor was not used anywhere so I've removed it. We can always add it back in with Claude if necessary later. I've checked that all the other constructors initialize all the data structures properly.

src/realm/activemsg.inl
lightsighter: I had Claude add support for this. See what you think.

src/realm/network.inl
muraj: This code doesn't compile if REALM_USE_MULTIPLE_NETWORKS is set, as there is no return if single_network == 0.  We should at least ensure this compiles correctly in this case.

src/realm/activemsg.h
```diff
@@ -170,6 +165,23 @@ namespace Realm {
     T *header;
     Realm::Serialization::FixedBufferSerializer fbs;
     uint64_t inline_capacity[INLINE_STORAGE / sizeof(uint64_t)];
```
muraj: Previous issue, but could be easily fixed here -- we should add `alignas(T)` here.

src/realm/activemsg.inl
```diff
@@ -550,70 +587,162 @@ namespace Realm {
...
-      uint64_t msg_id = next_message_id(target_);
-      send_single(0, 1, msg_id, payload_.data(), payload_.size());
-      return;
-    }
+      size_t offset = 0;
+      for(uint32_t chunk_id = 0; chunk_id < total_chunks; ++chunk_id) {
+        size_t chunk_size = std::min(max_chunk, total_payload - offset);
+
+        // dispatch to NodeSet or single-target constructor based on mode
+        if(chunk_targets_) {
```
muraj: We can clean this up by just defaulting to NodeSet for both unicast and multicast.  That should deduplicate a lot of this logic.

src/realm/activemsg.h
```diff
@@ -170,6 +165,23 @@ namespace Realm {
...
+
+  private:
+    // chunked-mode state: only used when the requested payload exceeds the
+    //  network's hard limit for a single message
+    size_t network_max_payload_{0}; // 0 = normal (non-chunked) mode
+    NodeID chunk_target_{0};
+    NodeSet *chunk_targets_{nullptr}; // allocated for multicast chunked mode
+    const void *chunk_src_data_{nullptr};
+    size_t chunk_src_datalen_{0};
+    char *chunk_alloc_{nullptr}; // owned buffer for network-allocated chunked mode
```
muraj: Would recommend `std::vector<char>`.

src/realm/activemsg.inl
```diff
@@ -550,70 +587,162 @@ namespace Realm {
...
+      assert(0 && "init_chunked_data called on WrappedWithFragInfo type");
+    } else {
+      // compute the per-chunk payload capacity using the wrapped header size
+      // pass the caller's data pointer so the backend can check registration
+      size_t wrapped_hdr_size = sizeof(WrappedWithFragInfo<T>);
+      network_max_payload_ = Network::max_payload_size(wrapped_hdr_size, _data);
+      assert(network_max_payload_ > 0);
+
+      chunk_target_ = _target;
+      chunk_src_data_ = _data;
```
muraj, line 644: This takes ownership of _data, which is probably not what we want.  Data passed here could go out of scope after creating the active message...

muraj (changes requested):

src/realm/activemsg.h
lightsighter: Fixed.

src/realm/activemsg.inl
lightsighter: Done.

src/realm/activemsg.h
lightsighter: I used `std::vector<std::byte>`.

src/realm/activemsg.inl
lightsighter: That is exactly what we want although we're not taking ownership. It is the caller's responsibility to maintain the liveness of the payload buffer until they call `commit` on the message. After the `commit` call is done then the caller has exclusive ownership of the buffer again and can do what they want with it (scribble in it, free it, etc). The `commit` call on an active message will either push messages into the networking layer ensuring the data is consumed by the network, or the networking backends themselves (UCX, GASNet) will make temporary copies of the data until they can push the active messages into the network. How that is done depends on the state of the networking backend though so we don't want to the `ActiveMessage` class making an unnecessary premature copy of the data.

src/realm/network.inl
lightsighter: What should we do to fix this? It should probably be its own separate pull request to fix all the functions and we can merge it and then upstream it to this branch.

src/realm/activemsg.inl
muraj: I see... it might be a good idea to document this behavior if it's not already, at least at the function declaration if not here because this really threw me.

src/realm/network.inl
muraj: At the very least, return 0 and assert(0) before it so it compiles and we can eval later.  This will allow it to compile and we can deal with the actual implementation later.  But it must be able to compile, right now it is not.

src/realm/activemsg.inl
lightsighter: I added a comment to this effect. There was already a comment there saying that these constructors have the semantics of `PAYLOAD_KEEP` which is what I described, but I further clarified it:
https://github.com/StanfordLegion/realm/blob/main/src/realm/activemsg.h#L79-L81

src/realm/network.inl
lightsighter: I put in calls to `std::abort()` in all the functions in that code path. It's better than `assert(0)` as it means you don't need to the `return 0` statement to keep the compiler happy since `std::abort` is marked as `[[noreturn]]`.

muraj (commented):
Looks good from the realm side.  Ready to approve assuming the network side of things looks good (UCX, etc), but I don't have the expertise to verify those paths.

src/realm/ucx/ucp_module.cc
```diff
@@ -333,4 +335,18 @@ namespace Realm {
         data, src_payload_addr.segment, &dest_payload_addr, with_congestion, header_size);
   }
 
+  size_t UCPModule::max_payload_size(size_t header_size, const void *src_payload_addr)
```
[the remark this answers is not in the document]
SeyedMir: That's a valid point. I suggest to update pbuf_get() to support any payload size. For sizes above the pre-registered mpool obj size, it will simply use malloc. I pushed a commit for the changes.

src/realm/ucx/ucp_module.cc
```diff
@@ -333,4 +335,13 @@ namespace Realm {
...
+  {
+    // caller-provided payloads are always copied into a pbuf before async
+    //  send, so the limit is the same regardless of src_payload_addr
```
SeyedMir: This statement is not correct. Caller-provided payloads may be sent using a zcopy protocol (large messages or GPU buffers are two examples).

src/realm/ucx/ucp_module.cc
```diff
@@ -333,4 +335,13 @@ namespace Realm {
...
+    (void)header_size;
+    (void)src_payload_addr;
+    return internal->config.pbuf_max_size;
```
SeyedMir: I think this can now always return SIZE_MAX (with the changes for pbuf_get that I pushed).

SeyedMir (commented):
Updated pbuf_get to make it support any payload size. Some other name changes were applied too to make them consistent and clear that they're mpool configs.

src/realm/ucx/ucp_module.cc
lightsighter: Ok, so we should delete this comment and return `SIZE_MAX` now?

src/realm/ucx/ucp_module.cc
lightsighter: Do you want to make this change or should I do it? I agree that the intention of this pull request is to make it so that we should never need to employee chunking on the Realm side for active message payloads when using the UCX network backend.

src/realm/ucx/ucp_module.cc
SeyedMir: Yes, pushed the commit.

src/realm/ucx/ucp_module.cc
SeyedMir: Made the changed and pushed.

> we should never need to employee chunking on the Realm side for active message payloads when using the UCX network backend.

Yes and no because depends on whether you consider the perf aspect too or not. I mean you will never need to chunk on Realm side for functionality (when using UCX), but you may still need to for perf. But, for that we already have recommended_max_payload() which the user should query for a more perf-aware return value.

lightsighter: > but you may still need to for perf

Can you describe a scenario when we may want to still employ chunking on the client side just so everyone is aware of what that might look like?

SeyedMir: Maybe not necessarily that Realm chunking will be better. But, I can think of two examples:

1. Sending a payload that is say 2x the size of pre-registered mpool buffers. In that case, you'll end up paying the registration cost on the payload buffer, whereas if you sent it in two chunks, then you would have avoided it. Now, the latter may not necessarily give you a better perf depending on the registration cost vs. chunking overheads.
2. If you decide to send a large payload without providing the remote address. In that case, max_payload_size will tell you "yes, I can send that" and what will happen is that you'll pay a malloc+registration cost on the receiver side too. But, having queried the recommended_max_payload with and without providing remote address would have given you different suggested sizes.

apryakhin: I think this PR is getting kind of messy from the history point of view by multiple people committing into it as well as doing AI-based implementation and then drastically changing it in between commits/reviews. I am not even sure what exactly is being checked- in at the moment from the last time I looked into it. Can someone please update the description to make sure it matches up with what's being done?

lightsighter: This comment is still up to date:
https://github.com/StanfordLegion/realm/pull/410#issuecomment-4057915685

muraj (approved):
Looks okay to me for the realm side.

SeyedMir:
LGTM.

elliottslaughter:
I ran this in combination with #428 and #433 and fuzzer has been happy with this change (multi-node debug + release, GASNet and UCX).

Note that for these initial runs, the fuzzer is not specifically configured to require fragmented AMs, so this basically confirms that there are no regressions for the base case. It doesn't necessarily prove anything about the AM fragmentation code otherwise.

elliottslaughter (changes requested):
I'm hitting fuzzer failures on this one, please hold until I can test the fragmentation path fully.

elliottslaughter (approved):
After fixing unrelated Legion issues, this passes my fuzzer runs with configurations wide enough to trigger some handful of fragmented messages (a couple dozen per run). I ran that configuration through my various options (debug/release, control replicated and not) and didn't hit any failures.

I'm happy enough with this for now. Once everyone else is on board feel free to merge.

src/realm/ucx/ucp_module.cc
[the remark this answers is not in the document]
eddy16112: @apryakhin I looked at the PR again. This issue has been fixed by setting the receiver buffer to `*buf = PBUF_TAG_MALLOC;`, so we won't get a fixed size object from the mpool, instead, we use malloc to allocate a new buffer.

elliottslaughter (commented):
I want to hold this a bit longer while we investigate https://github.com/StanfordLegion/legion/issues/1945. Current debugging efforts suggest that the livelock situation exhibited there is related in some fashion to this PR.

elliottslaughter (changes requested):

elliottslaughter (approved):
As discussed in the meeting today, I'm going to unblock this on my end. There's a fair argument that the current known fuzzer failure modes are unrelated to this, and at any rate they're almost exceedingly rare, while this PR resolves very clearly demonstrated issues for Legion users.

Feel free to merge once everyone else is satisfied.

eddy16112 (approved):

PR #419, opened 2026-04-07
Copy Planning Timeout

lightsighter:
Summary                                                                                                                               
   
  - Modified TransferDesc::perform_analysis to accept a TimeLimit parameter and return a bool indicating whether analysis completed (true) or timed out (false)                                                                                                         
  - Added incremental state members (analysis_init_done, analysis_field_idx, analysis_fld_start, analysis_fill_ofs, analysis_field_done) to TransferDesc so that a timed-out analysis can be resumed from where it left off without recomputing prior work                    
  - Pre-loop initialization (profiling setup, fill data allocation, dimension ordering, field resizing) is guarded by analysis_init_done and only runs once across calls                                                                                                      
  - The main field loop checks work_until.is_expired() at each iteration, saving progress and returning false if the time limit is reached                                                                                                                               
  - Both existing callsites updated: check_analysis_preconditions passes an unbounded TimeLimit(), and DeferredAnalysis::event_triggered forwards its work_until parameter (previously ignored per the // TODO: respect time limit comment)

### DIFF
--- a/src/realm/transfer/transfer.cc
+++ b/src/realm/transfer/transfer.cc
@@ -3534,6 +3534,19 @@ namespace Realm {
     }
   }
 
+  TransferDesc::TransferDesc(TestTag, TransferDomain *_domain)
+    : refcount(1)
+    , deferred_analysis(this)
+    , prs()
+    , analysis_complete(false)
+    , analysis_successful(false)
+    , fill_data(0)
+    , fill_size(0)
+    , analysis_init_done(false)
+  {
+    domain = _domain;
+  }
+
   void TransferDesc::check_analysis_preconditions()
   {
     log_xplan.info() << "created: plan=" << (void *)this << " domain=" << *domain
--- a/src/realm/transfer/transfer.cc
+++ b/src/realm/transfer/transfer.cc
@@ -3609,7 +3609,7 @@ namespace Realm {
     }
 
     // no (untriggered) preconditions, so we fall through to immediate analysis
-    perform_analysis();
+    perform_analysis(TimeLimit());
   }
 
   static size_t compute_ib_size(size_t combined_field_size, size_t domain_size,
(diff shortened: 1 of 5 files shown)


src/realm/transfer/transfer.h
```diff
@@ -456,6 +464,11 @@ namespace Realm {
     std::vector<FieldInfo> src_fields, dst_fields;
     void *fill_data;
     size_t fill_size;
+    bool analysis_init_done;
```
apryakhin: nit: all this members are uninitialized

src/realm/transfer/transfer.cc
```diff
@@ -3678,65 +3691,81 @@ namespace Realm {
...
-    std::vector<bool> field_done(srcs.size(), false);
+    size_t fill_ofs = analysis_fill_ofs;
+    size_t fld_start = analysis_fld_start;
+    auto &field_done = analysis_field_done;
     // fields will get reordered to be contiguous per xd subgraph
-    size_t fld_start = 0;
-    for(size_t i = 0; i < srcs.size(); i++) {
+    for(size_t i = analysis_field_idx; i < srcs.size(); i++) {
+      // check time limit between field iterations
+      if(work_until.is_expired()) {
```
apryakhin, line 4114: The work_until check is only inside the field loop, but the init phase (`choose_dim_order`, fill allocation, field resizing) runs unconditionally before it. A tight or expired time limit won't be caught until the loop starts.

More critically, the `TransferGraph` is a local variable - destroyed on timeout, but already-processed fields are skipped on resume via `field_done`. The resumed graph is incomplete. This needs to be a `TransferDesc` member to work correctly.

I suggest either making this a pure signature change (document that incremental resume isn't implemented yet) or finishing the implementation. 

The current state looks functional but isn't really.

lightsighter: > The work_until check is only inside the field loop, but the init phase (choose_dim_order, fill allocation, field resizing) runs unconditionally before it. A tight or expired time limit won't be caught until the loop starts.

I think these are all relatively cheap and can probably be done without checking the timeout. Choosing the dimension order is always a small integer constant loop. Fill allocation just requires one loop over the fields with an addition for each field. The resizing requires at most a handful of mallocs with linear initialization which will all hit in the L1 cache.

> More critically, the TransferGraph is a local variable - destroyed on timeout, but already-processed fields are skipped on resume via field_done. The resumed graph is incomplete. This needs to be a TransferDesc member to work correctly.

The `TransferGraph` is a member of the `TransferDesc` here:
https://github.com/StanfordLegion/realm/blob/mbauer-planning-timeout/src/realm/transfer/transfer.h#L462
so its state should persist across the calls.

src/realm/transfer/transfer.h
lightsighter: They're all initialized in the base `TransferDesc` constructor here:
https://github.com/StanfordLegion/realm/blob/mbauer-planning-timeout/src/realm/transfer/transfer.cc#L3543-L3545
That constructor is called by all the other constructors.

src/realm/transfer/transfer.cc
```diff
@@ -4294,7 +4324,7 @@ namespace Realm {
     if(poisoned) {
       desc->cancel_analysis(precondition);
     } else {
-      desc->perform_analysis();
+      desc->perform_analysis(TimeLimit());
```
eddy16112, line 4690: pass the `work_until` to the `perform_analysis`?

src/realm/transfer/transfer.h
```diff
@@ -29,6 +29,8 @@
 #include "realm/transfer/channel.h"
 #include "realm/profiling.h"
 
+class PerformAnalysisTest;
```
eddy16112: I do not like the idea of exposing the testing class here, but I do not have any other better solutions.

src/realm/transfer/transfer.h
eddy16112: The analysis_init_done is initialized in the constructor, but analysis_field_idx, analysis_fld_start and analysis_fill_ofs are not. IMHO, I would prefer to initialize them here, it is easier for the readers to find out the default values.

eddy16112 (commented):
We never check the return value of `perform_analysis` in this PR?

src/realm/transfer/transfer.cc
```diff
@@ -3534,6 +3534,19 @@ namespace Realm {
...
+    : refcount(1)
+    , deferred_analysis(this)
+    , prs()
+    , analysis_complete(false)
+    , analysis_successful(false)
+    , fill_data(0)
+    , fill_size(0)
+    , analysis_init_done(false)
+  {
+    domain = _domain;
```
eddy16112: why not do `domain(_domain)`?

lightsighter:
> We never check the return value of `perform_analysis` in this PR?

That's right because we always pass in "infinite time" for perform analysis so it will never not complete. It's only after we merge this pull request into the `main` branch and I upstream it in the background-work planning branch that this code will start to be used.

src/realm/transfer/transfer.cc
lightsighter: That would result in incorrect code currently because we don't have any machinery in this branch to defer it if it times out. We can't make this change until this branch is merged and it is upstreamed into the background copy planning branch.

src/realm/transfer/transfer.h
lightsighter: It doesn't cost us anything to do have a forward declaration for this testing class and it was necessary to get access to the internals of the transfer descriptor for having a unit test to satisfy the required by the CI that we have test coverage of all the new lines. Is there another way that you prefer to do this kind of unit testing elsewhere in Realm?

src/realm/transfer/transfer.h
lightsighter: Ok, I just added the default initializers in the class definition.

src/realm/transfer/transfer.cc
lightsighter: Fixed.

src/realm/transfer/transfer.h
eddy16112: One way I can think of is to create a child class from `TransferDesc` and expose the `perform_analysis` as public:
```
class TestableTransferDesc : public TransferDesc {
public:                                         
  using TransferDesc::perform_analysis;        
};
```
Now, we are able to use `TestableTransferDesc` in the unit test.

lightsighter: Good suggestion, done!

eddy16112 (approved):

muraj (approved):

src/realm/transfer/transfer.cc
lightsighter: I'm going to go ahead and resolve this as I believe the functionality is correct. If we want to revisit how frequently we do the testing of the timeout that is easy to do later.

PR #423, opened 2026-04-24
Fix Fast Reservations Again

lightsighter:
## Summary                                                                                                                    
   
  Fixes a missed-wakeup hang where `rdlock_slow`/`wrlock_slow` returned a `waiter_event` to the caller that was never triggered, leaving the thread stuck waiting on a lock whose state had returned to 0.                                                    
  
  ## Root cause                                                                                                                 
                                                            
  The invariant the code relied on — "if `frs.waiter_event` is set, then `STATE_WRITER_WAITING` is set in state, so the next    
  unlock will go through `unlock_slow` and trigger the event" — was broken in two places:
                                                                                                                                
  1. **`wrlock_slow` line 948**: `CAS(WW, W)` acquired the lock by transitioning state from `WW` to `W`, clearing               
  `STATE_WRITER_WAITING` as a side effect. If `frs.waiter_event` was set at that moment, the acquiring writer's subsequent fast-path unlock would succeed (WW no longer set), bypass `unlock_slow`, and never trigger the event.                         
                                                            
  2. **Orphan cleanup in case 3/4 `else` branch**: cleared `STATE_WRITER_WAITING` via `fetch_and(~WW)` without checking `frs.waiter_event`. Same orphan scenario.                                                                                     
  
  ## Changes                                                                                                                    
                                                            
  ### `src/realm/rsrv_impl.cc`
                                                                                                                                
  - Changed `CAS(WW, W)` → `CAS(WW, W|WW)` at line 948. Writers acquiring from a WW-set state now preserve WW, forcing their subsequent unlock through `unlock_slow` where `waiter_event` gets triggered.                                                  
  - Orphan cleanup now triggers `frs.waiter_event` if set, before clearing `STATE_WRITER_WAITING`. Applies to both `wrlock_slow` and `rdlock_slow`.                                                                                                           
  - Updated WW-set loop comment: with the `CAS(WW, W|WW)` change, the loop's third branch (WW cleared while lock held) is no longer reachable, so the loop terminates in 1 iteration. Kept the bounded loop with abort as defense-in-depth.                
  - Added `REALM_ASSERT((cur_state & STATE_BASE_RSRV) == 0)` before each `fetch_sub(BRW - BR)` call site (3 sites) to catch protocol violations that would corrupt the reader count. Added `#include "realm/realm_assert.h"`.                             
                                                            
  ## Correctness                                                                                                                
                                                            
  With these changes, the invariant
                                                                                                                                
  > If `frs.waiter_event` is set, then `STATE_WRITER_WAITING` is set in state (or a thread holding `frs.mutex` will establish 
  one of those atomically before releasing) now holds across every state mutation. Every path that clears `STATE_WRITER_WAITING` either preserves it (the `CAS(WW, W|WW)` change) or triggers `frs.waiter_event` if set (all other paths: `unlock_slow`, `advise_sleep_entry`, WW-set loop first branch, orphan cleanup).
                                                                                                                                
  ## Performance                                            

  - Uncontended operations: zero overhead.
  - Contended operations with a registered waiter: zero overhead (slow-path unlock was already required).
  - Edge case (writer acquires from `state=WW` with no actual waiter): one extra `frs.mutex` round-trip on unlock, ~100-200ns.  
  Rare in practice.                                                                                                             
                                                                                                                                
  ## Testing                                                                                                                    
                                                            
  716/716 unit tests pass.

### DIFF
--- a/src/realm/rsrv_impl.cc
+++ b/src/realm/rsrv_impl.cc
@@ -21,6 +21,7 @@
 #include "realm/event_impl.h"
 #include "realm/runtime_impl.h"
 #include "realm/timers.h"
+#include "realm/realm_assert.h"
 
 namespace Realm {
 
@@ -944,7 +945,12 @@ namespace Realm {
       if((cur_state & (STATE_SLOW_FALLBACK | STATE_BASE_RSRV | STATE_BASE_RSRV_WAITING |
                        STATE_SLEEPER)) == 0) {
         State prev_state = (cur_state & STATE_WRITER_WAITING);
-        State new_state = STATE_WRITER;
+        // preserve WRITER_WAITING when acquiring - if a waiter_event is
+        //  registered, we need the unlock to go through the slow path so
+        //  the event gets triggered.  Without this, acquiring via
+        //  CAS(WW, WRITER) would clear WW and the next unlock would take
+        //  the fast path, orphaning the waiter_event.
+        State new_state = STATE_WRITER | prev_state;
         if(state.compare_exchange(prev_state, new_state))
           return Event::NO_EVENT;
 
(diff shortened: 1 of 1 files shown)


elliottslaughter:
Fuzzer completed successfully with only known failure modes. No hangs, no memory corruption. MutexChecker failure is gone. The only failure mode is "network not quiescent", which is not expected to be fixed by this or any other recent commit.

elliottslaughter (approved):

elliottslaughter:
Legion CI also passed a full CI run on this branch.

PR #424, opened 2026-04-26
Fix for GASNetEX Quiescence Checking

lightsighter:
Fix GASNet-EX quiescence check spurious failures during shutdown                                                              
                                                                                                                                
  Background                                                                                                                    
                                                                                                                                
  During runtime shutdown, RuntimeImpl::wait_for_shutdown calls Network::check_for_quiescence in a loop (up to 10 attempts) to confirm that no messages remain in flight before tearing down the network. CI runs against the GASNet-EX backend frequently fail this loop with:                                                                                                          
                                                            
  network still not quiescent after 10 attempts
                                                                                                                                
  The same CI configuration running against the UCX backend on the same machines does not exhibit this failure, indicating the issue is specific to the GASNet-EX backend rather than a fundamental problem with the quiescence algorithm or the iteration limit.                                                                                                                        
                                                            
  Root cause

  The quiescence check works by allreducing per-rank counters (sum of packets_reserved across xpairs, total_packets_received) and requiring globally-balanced sends/receives plus all "busy" indicators to be zero. The algorithm is structurally identical across the GASNet-EX, UCX, and MPI backends.                                                                                  
                                                            
  Where the backends differ is in how they ensure progress is made between consecutive iterations of the retry loop:            
  
  - MPI (mpi_module.cc:836) calls g_am_manager.ensure_polling_progress() at the start of every check — guaranteeing a full poll cycle has completed before the next snapshot is taken.    
  - UCX (ucp_internal.cc:1748-1755) calls poller.wait_polling() on every poller at the end of any check that returns false —    
  same guarantee, applied symmetrically on the failure path.                                                                    
  - GASNet-EX calls poller.wait_for_full_poll_cycle() only during the gex_coll_ireduce wait loop. If the collective happens to
  complete on the first iteration of that loop, the check returns after a single gex_am_poll — and that poll happened before the result was even known, so it cannot react to "this iteration was not stable, give the system time to settle." The runtime loop then immediately calls check_for_quiescence again with no enforced settle time.                                          
                                                            
  The result: on a fast collective, GASNet-EX gets effectively zero between-iteration settle cycles, so transient activity (in-flight completion replies, reverse-get RDMAs landing, handler-driven sends) keeps the snapshots permanently out of sync until the 10-attempt budget is exhausted. UCX and MPI structurally avoid this because their settle is unconditional.          
                                                            
  This is a backend-local issue and should be fixed locally, not by changing the cross-backend Network::check_for_quiescence interface.
                                                                                                                                
  Changes                                                   

  All four edits live in src/realm/gasnetex/gasnetex_internal.cc:

  1. GASNetEXInternal::check_for_quiescence (~L3652-3661) — On the "not quiescent" branch, call poller.wait_for_full_poll_cycle() once more before returning false. This mirrors UCX's wait_polling() placement and provides the same between-iteration settle guarantee that MPI and UCX rely on.                                                         
  2. GASNetEXPoller::end_polling (~L2792-2795) — Clear started = false under the mutex after the poller has confirmed shutdown. This makes the post-shutdown state observable to safety checks.                                                               
  3. GASNetEXPoller::wait_for_full_poll_cycle (~L2980-2983) — Assert started under the existing mutex acquisition. If a future caller invokes check_for_quiescence after the poller has been shut down, the assertion fires immediately rather than deadlocking on pollwait_cond (which would never be signalled, since the poller is no longer being requeued by bgwork).
  4. GASNetEXInternal::detach (~L3462-3465) — Document the invariant near poller.end_polling(): no check_for_quiescence calls past this point. This guards against future regressions if someone adds a new caller of the quiescence check that doesn't follow the existing shutdown ordering.
                                                                                                                                
  Why this is the right fix                                 

  - No interface change. The asymmetry was a gasnetex-local oversight; UCX and MPI already do the right thing. Modifying        
  Network::check_for_quiescence's signature to thread totals or stability state through the caller would be churn for a problem only one backend has.                                                                                                         
  - No timing constants. The wait is on a poll-cycle event, not wall-clock time, so there is nothing to tune per machine.
  - Mirrors a known-good pattern. The fix brings GASNet-EX in line with the settle-cycle guarantee that UCX and MPI already provide.                                                                                                                      
  - Defensive against future regressions. The added assertion catches any future violation of the "no quiescence checks past detach" invariant immediately rather than presenting as a hang.                                                               
                                                            
  Static analysis of safety                                                                                                     
                                                            
  The new wait_for_full_poll_cycle call site can only fire from RuntimeImpl::wait_for_shutdown (the sole caller of Network::check_for_quiescence in the source tree), strictly before network_modules->detach() runs. At that point in the shutdown sequence, begin_polling has been called, shutdown_flag is false, and the poller is still being requeued by bgwork —  all preconditions for wait_for_full_poll_cycle to make forward progress are satisfied. The retry loop's two exits (success → break to detach; failure → abort) are unaffected.

  Testing

  The fix should be validated by running the CI configurations that have been seeing the "not quiescent after 10 attempts" failure — primarily the GASNet-EX matrix. A clean run confirms the fix; a continued failure would indicate the residual cause is something a single poll cycle cannot resolve (e.g., work that needs the injector or rgetter to advance independently), in which case a follow-up change to widen the wait or move to a progress-based retry loop would be appropriate.

### DIFF
--- a/src/realm/bgwork.cc
+++ b/src/realm/bgwork.cc
@@ -389,22 +389,24 @@ namespace Realm {
                        << " slot=" << index;
 #ifdef DEBUG_REALM
     State old_state = state.exchange(STATE_ACTIVE);
-    if(old_state != STATE_IDLE) {
-      log_bgwork.debug() << "double make_active: item=" << ((void *)this) << " name='"
-                         << name << "' oldstate=" << old_state;
+    if(old_state == STATE_SHUTDOWN) {
+      log_bgwork.fatal() << "make_active on shutdown item: item=" << ((void *)this)
+                         << " name='" << name << "'";
+      abort();
     }
 #endif
     manager->advertise_work(index);
   }
 
 #ifdef DEBUG_REALM
-  // called immediately before 'do_work'
+  // called immediately before 'do_work' - may observe STATE_IDLE if
+  //  an external make_active raced with another worker's make_inactive
   void BackgroundWorkItem::make_inactive(void)
   {
     State old_state = state.exchange(STATE_IDLE);
-    if(old_state != STATE_ACTIVE) {
-      log_bgwork.fatal() << "invalid make_inactive: item=" << ((void *)this) << " name='"
-                         << name << "' oldstate=" << old_state;
+    if(old_state == STATE_SHUTDOWN) {
+      log_bgwork.fatal() << "make_inactive on shutdown item: item=" << ((void *)this)
+                         << " name='" << name << "'";
       abort();
     }
   }
--- a/src/realm/bgwork.h
+++ b/src/realm/bgwork.h
@@ -154,8 +154,10 @@ namespace Realm {
 
 #ifdef DEBUG_REALM
   public:
-    // in debug mode, we'll track the state of a work item to avoid
-    //  duplicate activations or activations after shutdown
+    // in debug mode, we'll track the state of a work item to detect
+    //  activations or work performed after shutdown - duplicate
+    //  activations are permitted and may occur when an external thread
+    //  calls make_active while a worker is between claim and do_work
     enum State
     {
       STATE_IDLE,
--- a/src/realm/gasnetex/gasnetex_internal.cc
+++ b/src/realm/gasnetex/gasnetex_internal.cc
@@ -2789,6 +2789,10 @@ namespace Realm {
       assert(!shutdown_flag.load());
       shutdown_flag.store(true);
       shutdown_cond.wait();
+      // poller is no longer being requeued by bgwork - clear 'started' so
+      //  any future call to wait_for_full_poll_cycle() trips an assertion
+      //  rather than deadlocking on pollwait_cond
+      started = false;
     }
   }
 
@@ -2974,6 +2978,9 @@ namespace Realm {
   void GASNetEXPoller::wait_for_full_poll_cycle()
   {
     AutoLock<> al(mutex);
+    // the poller must still be scheduled by bgwork - otherwise nothing will
+    //  ever clear pollwait_flag and the wait below would block forever
+    assert(started);
     pollwait_flag.store(true);
     pollwait_cond.wait();
   }
@@ -3452,6 +3459,10 @@ namespace Realm {
 
   void GASNetEXInternal::detach()
   {
+    // NOTE: check_for_quiescence() must not be called after this point -
+    //  end_polling() stops the poller from being requeued by bgwork, which
+    //  would cause wait_for_full_poll_cycle() inside the quiescence check
+    //  to deadlock (an assert in wait_for_full_poll_cycle catches violations)
     poller.end_polling();
 
     // drain any xpairs that were enqueued (via request_push) but never
@@ -3638,7 +3649,17 @@ namespace Realm {
       log_gex_quiesce.debug() << "total counts: " << total_counts[0] << " "
                               << total_counts[1] << " " << total_counts[2];
 
-    return ((total_counts[0] == 0) && (total_counts[1] == total_counts[2]));
+    bool quiescent = ((total_counts[0] == 0) && (total_counts[1] == total_counts[2]));
+
+    // if we're not quiescent, force at least one more full poll cycle before
+    //  returning so the caller's next snapshot can't outrun the poller - the
+    //  ireduce above only guarantees one poll cycle if it completed slowly,
+    //  and a fast collective leaves the system with no settle time between
+    //  retries (unlike UCX's wait_polling and MPI's ensure_polling_progress)
+    if(!quiescent)
+      poller.wait_for_full_poll_cycle();
+
+    return quiescent;
   }
 
   PendingCompletion *GASNetEXInternal::get_available_comp()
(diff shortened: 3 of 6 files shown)


elliottslaughter:
Copying message from Zulip for posterity. I'm still seeing network not quiescent with this PR:

```
(gdb) bt
#0  0x00007f088c45c23f in __GI___clock_nanosleep (clock_id=clock_id@entry=0, flags=flags@entry=0, req=req@entry=0x7fff9cec1060, rem=rem@entry=0x7fff9cec1060) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
#1  0x00007f088c461ec7 in __GI___nanosleep (requested_time=requested_time@entry=0x7fff9cec1060, remaining=remaining@entry=0x7fff9cec1060) at nanosleep.c:27
#2  0x00007f088c461dfe in __sleep (seconds=0, seconds@entry=1) at ../sysdeps/posix/sleep.c:55
#3  0x00007f088c9d1172 in Realm::realm_freeze (signal=<optimized out>) at /scratch/eslaught/fuzzer-experiment-2026-04-26-mr2042-realm-pr424/legion/realm/src/realm/runtime_impl.cc:238
#4  <signal handler called>
#5  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#6  0x00007f088c3a1859 in __GI_abort () at abort.c:79
#7  0x00007f088c9d5181 in Realm::RuntimeImpl::wait_for_shutdown (this=<optimized out>) at /scratch/eslaught/fuzzer-experiment-2026-04-26-mr2042-realm-pr424/legion/realm/src/realm/runtime_impl.cc:2902
#8  0x00007f088c9d5c33 in Realm::Runtime::wait_for_shutdown (this=this@entry=0x7fff9cec2390) at /scratch/eslaught/fuzzer-experiment-2026-04-26-mr2042-realm-pr424/legion/realm/src/realm/runtime_impl.cc:706
#9  0x00007f088de4ae28 in Legion::Internal::Runtime::start (argc=<optimized out>, argv=<optimized out>, background=<optimized out>, supply_default_mapper=<optimized out>, filter=<optimized out>, init_network=<optimized out>, configure_machine=true)
    at /scratch/eslaught/fuzzer-experiment-2026-04-26-mr2042-realm-pr424/legion/runtime/legion/kernel/runtime.cc:10956
#10 0x000055aa22ec5c29 in main (argc=<optimized out>, argv=<optimized out>) at /scratch/eslaught/fuzzer-experiment-2026-04-26-mr2042-realm-pr424/src/fuzzer.cc:1334
(gdb) frame 7
#7  0x00007f088c9d5181 in Realm::RuntimeImpl::wait_for_shutdown (this=<optimized out>) at /scratch/eslaught/fuzzer-experiment-2026-04-26-mr2042-realm-pr424/legion/realm/src/realm/runtime_impl.cc:2902
2902              abort();
(gdb) list
2897            }
2898
2899            if(tries >= 10) {
2900              log_runtime.fatal() << "network still not quiescent after " << tries
2901                                  << " attempts";
2902              abort();
2903            }
2904          }
2905        }
2906
```

```
commit fb6968b98a14af2524395dfc25d4dac4e6a9cac0 (HEAD -> test-mr2042-plus-realm-pr424, origin/test-mr2042-plus-realm-pr424)
Merge: 208e66783 5b12eb8ce
Author: Elliott Slaughter <slaughter@cs.stanford.edu>
Date:   Sun Apr 26 07:52:23 2026 -0700

    Merge commit '5b12eb8ce7d6ebc4251f9e163ba121e4fc878f95' into test-mr2042-plus-realm-pr424

commit 5b12eb8ce7d6ebc4251f9e163ba121e4fc878f95
Author: Elliott Slaughter <slaughter@cs.stanford.edu>
Date:   Sun Apr 26 07:52:23 2026 -0700

    Squashed 'realm/' changes from 77e872a5a9..726629aa6f
    
    726629aa6f realm: proposed fix for gasnetex quiescence checking
    d9c0cb75e5 Fix Fast Reservations Again (#423)
    c271dec00a Copy Planning Timeout (#419)
    d393ba4e3a Fix Event Triggering Race (#420)
    30da2be37b Fix GASNet Races (#409)
    442cab573a Support transfers for layouts with multiple id-indexed fields (#284)
    f3373796cc Sm/cuda redop opt (#330)
    87ec0ada55 Spin Lock Fixed 3 (#415)
    1b9041582d Fix Spin Locks 2 (#414)
    c178f8242e Fix Reduction Tutorial Race (#411)
    c28d3f9932 Fix Spin Lock Hang (#407)
    ea52777415 Fix Missed Wakeup in Remote Processor Group Spawns (#408)
    fa0f8685b7 Fix the redop path when p2p is disable (#402)
    
    git-subtree-dir: realm
    git-subtree-split: 726629aa6f73ea2b9aa30cf8ceb91b8c5d330d52
```

lightsighter:
I had claude add some diagnostic logging so we can better understand why quiescence isn't being achieved. Let's try running again with the latest version of this branch and see what logging gets printed.

lightsighter:
Closing this in favor of #428

PR #425, opened 2026-04-27
More Fast Reservation Fixes

lightsighter:
This is an incremental fix on top of the earlier `Fix Fast Reservations Again (#423)` work. It addresses three remaining bugs found via fuzzer crashes and code review.
                                                                                                                                
  ## Bugs fixed                                             

  ### 1. Use-after-free in `unlock_slow`                                                                                        
   
  The previous `unlock_slow` did its final `state.fetch_sub_acqrel(...)` (releasing the public lock) before `frs.mutex.unlock()`. After the public lock release, an external owner (e.g., a Legion `DistributedCollectable` observing the lock-free state) could deallocate the `FastReservation`. The subsequent `frs.mutex.unlock()` then operated on freed/reused memory, manifesting as `UnfairMutex` assertion failures and accumulating reader-count corruption.                             
                                                            
  ### 2. Orphaned `waiter_event` during the release window                                                                      
   
  After re-ordering `unlock_slow` to release the mutex before the public lock, a window opens where a new mutex-path entrant could take `frs.mutex`, create a new `waiter_event`, set `STATE_WRITER_WAITING`, and return waiting on it — only to be orphaned because the original lock holder's `state.fetch_sub` was about to clear the lock without triggering the new event.
                                                                                                                                
  ### 3. Missing BRW handler in `wrlock_slow`'s case switch (latent)                                                            
   
  `wrlock_slow`'s mutex-path case switch handled BR (case 1), SLEEPER (case 2), and clean R/W/WW (case 3), but had no handler for `STATE_BASE_RSRV_WAITING` set without BR or SLEEPER. `rdlock_slow` had this case (its case 2). When a writer's slow path encountered the BRW-only state — reachable when a remote node demands the base reservation back during writer contention — it hit the fatal "unexpected state" assert. Recently observed via fuzzer (likely also exposed indirectly by the corruption from bug 1, but is a latent bug regardless).                   

  ## Changes

  ### `src/realm/reservation.h`                                                                                                 
   
  - Added `STATE_RELEASING = 0x02000000` (bit 25). Reduced `STATE_READER_COUNT_MASK` to 25 bits (`0x01ffffff`, max ~33M concurrent readers).                                      
                                                                                                                                
  ### `src/realm/rsrv_impl.cc`                              

  **`unlock_slow` restructured to fix the UAF:**
                                                                                                                                
  - Releases `frs.mutex` BEFORE the final `state.fetch_sub_acqrel(...)`. All `frs.*` accesses (including `frs.mutex.unlock()`) now complete while the public lock is still held, ensuring the FastReservation is alive throughout. After the final           
  `fetch_sub_acqrel`, no `frs.*` access occurs.             
  - Snapshots `frs.waiter_event` under the mutex before the release; triggers it after `frs.mutex.unlock()` and `fetch_sub` (using only the local `to_trigger` UserEvent, not any `frs.*` field).                                                         
  - Reader path: non-last readers decrement under the mutex (preserves serialization across concurrent reader unlocks); only the last reader defers the decrement until after `frs.mutex.unlock()`.                                                           
                                                            
  **`STATE_RELEASING` mechanism to close the new-waiter orphan window:**                                                        
                                                                                                                                
  - `unlock_slow` sets `STATE_RELEASING` under the mutex via `state.fetch_or(STATE_RELEASING)`, then releases the mutex. New mutex-path entrants see `STATE_RELEASING` and spin instead of creating a `waiter_event`. `STATE_RELEASING` is cleared atomically with `STATE_WRITER` (or last reader's R count) in the final `fetch_sub_acqrel`, along with `STATE_WRITER_WAITING` if it was set.                                                                                                                
  - Uses the return value of `state.fetch_or(STATE_RELEASING)` (the pre-set state) to compute `bits_to_clear`, capturing any `STATE_WRITER_WAITING` that a spinner may have set between the initial `cur_state` load and our `fetch_or`. After `STATE_RELEASING` is set, no more spinner CASes can succeed.
  - Both `wrlock_slow` and `rdlock_slow` outer loops check `STATE_RELEASING` at the top of each iteration and spin if set.      
  - Both slow paths have a defensive secondary `STATE_RELEASING` check after acquiring `frs.mutex` (in case `STATE_RELEASING` was set between the outer-loop check and mutex acquisition).                                                                  
  - `trywrlock_slow` treats both `STATE_RELEASING` and `STATE_BASE_RSRV_WAITING` as contention (returns false) in both the outer fast-path check and the post-mutex-acquisition check.                                                                        
                                                            
  **`wrlock_slow` BRW handler:**                                                                                                
                                                                                                                                
  - Added a new case 2 to `wrlock_slow`'s mutex-path case switch (mirroring `rdlock_slow`'s case 2): handles 
  `STATE_BASE_RSRV_WAITING` by either swapping BRW→BR + releasing the base reservation (when no holders) or just calling `request_base_rsrv` to register a pending request, then returning the resulting event. Includes the `REALM_ASSERT((cur_state & STATE_BASE_RSRV) == 0)` guard before the BRW-swap arithmetic.                                                                
  - Renumbered the subsequent cases (old case 2 SLEEPER → case 3; old case 3 clean R/W/WW → case 4) for consistency with `rdlock_slow`.                                                                                                                
   
  ## Correctness                                                                                                                
                                                            
  Five-property static analysis was conducted to verify the fix:
                                                                                                                                
  1. **State-machine correctness**: every reachable `cur_state` at every program point now has a handler. The previous BRW gap in `wrlock_slow` is closed; `STATE_RELEASING` states are filtered by the outer-loop and mutex-path-entry checks before reaching the case switch.                                 
  2. **Invariant preservation**: 11 named invariants verified across every primitive state mutation, including new invariants I9 (RELEASING ⟹ W or R>0), I10 (at most one thread sets RELEASING at a time), I11 (state stable while RELEASING is set, except by the setter's `fetch_sub`).
  3. **Lifetime / memory safety**: every `frs.*` access (including `frs.mutex.unlock()`) happens while the FastReservation is   
  guaranteed alive (the public lock is still held).                                                                             
  4. **Deadlock freedom**: no lock-order cycles introduced; no re-entrancy issues; bounded RELEASING-spin.
  5. **Liveness**: the `STATE_RELEASING` mechanism eliminates the orphan-event hang that would otherwise be possible during the release window.                                                                                                               
                                                                                                                                
  ## Performance                                                                                                                
                                                            
  All changes are confined to slow paths or are no-op-equivalent constant changes:
                                                                                                                                
  - **Uncontended operations: zero overhead.**
  - **Contended operations with a registered waiter: zero overhead.**                                                           
  - **Reader count maximum reduced** from ~67M to ~33M (one bit reallocated to `STATE_RELEASING`). Realistic workloads are far below this limit.

### DIFF
--- a/src/realm/reservation.h
+++ b/src/realm/reservation.h
@@ -129,7 +129,8 @@ namespace Realm {
     // the fast path stores several things in a 32-bit word that can be
     //  atomically updated
     typedef uint32_t State;
-    static const State STATE_READER_COUNT_MASK = 0x03ffffff;
+    static const State STATE_READER_COUNT_MASK = 0x01ffffff;
+    static const State STATE_RELEASING = 0x02000000;
     static const State STATE_SLEEPER = 0x04000000;
     static const State STATE_WRITER = 0x08000000;
     static const State STATE_WRITER_WAITING = 0x10000000;
--- a/src/realm/rsrv_impl.cc
+++ b/src/realm/rsrv_impl.cc
@@ -938,6 +938,17 @@ namespace Realm {
       // read the current state to see if any exceptional conditions exist
       State cur_state = state.load_acquire();
 
+      // if the previous lock holder is in the middle of releasing the lock,
+      //  spin briefly until they finish.  We must not enter the mutex path
+      //  in this case because creating a new waiter_event would orphan it
+      //  (the releasing thread has already done its waiter_event handling).
+      //  STATE_RELEASING is only set transiently between the releaser's
+      //  mutex.unlock and their final state.fetch_sub.
+      if((cur_state & STATE_RELEASING) != 0) {
+        REALM_SPIN_YIELD();
+        continue;
+      }
+
       // if there are no exceptional conditions (sleepers, base_rsrv stuff),
       //  try to clear the WRITER_WAITING (if set), set WRITER, on the
       //  assumption that the READER_COUNT is 0 (i.e. we want the CAS to fail
@@ -987,6 +998,16 @@ namespace Realm {
         //  cannot change out from under us
         cur_state = state.load_acquire();
 
+        // STATE_RELEASING may have been set by a previous holder's
+        //  unlock_slow between our outer-loop check and our mutex
+        //  acquisition. If it's set, release the mutex and spin in the
+        //  outer loop instead of creating a waiter_event.
+        if((cur_state & STATE_RELEASING) != 0) {
+          frs.mutex.unlock();
+          REALM_SPIN_YIELD();
+          continue;
+        }
+
         // goal is to find (or possibly create) a condition we can wait
         //  on before trying again
         Event wait_for = Event::NO_EVENT;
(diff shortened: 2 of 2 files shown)


elliottslaughter (approved):
Passed two full fuzzer runs (in combination with https://gitlab.com/StanfordLegion/legion/-/merge_requests/2043), so I'm happy with this.

PR #426, opened 2026-04-29
Fix Background Work State Machine

lightsighter:
Fix a latent bug in checking the state of bgworker items


muraj (approved):
lgtm, thanks!

PR #428, opened 2026-04-30
Alternative Network Quiescence Algorithm

lightsighter:
Background                                                                                                                                            
   
  Quiescence detection during shutdown is currently failing intermittently in CI on both the GASNet-EX and UCX backends with network still not quiescent after 10 attempts. Investigation revealed three holes in the existing algorithm:
                                                                                                                                                        
  1. Timing-sensitive busy flags (GASNet-EX). injector::has_work_remaining() includes an "is a worker currently inside push_packets" condition, which can be true while the system is actually quiescent — for example, when a single GASNet send takes hundreds of microseconds to return even though no work is in flight or queued. Verified empirically via gdb on a frozen failing process: injector.in_progress=true was reported for all 10 quiescence iterations, but at gdb-attach time packets_reserved == packets_sent, all queues empty, compmgr.num_pending=0 — i.e., the system was actually quiescent the moment the long send returned.
  2. Fixed 10-attempt abort budget (all backends). The shutdown loop in runtime_impl.cc aborts after exactly 10 quiescence attempts regardless of whether progress is being made. Each iteration runs in well under 100µs, so the entire 10-attempt window is shorter than a single slow GASNet/UCX send. This makes the algorithm fail under any non-trivial load even when the underlying state machine is sound.
  3. Single-round stability check (all backends). Each backend's previous check was a single-round vote with no cross-iteration confirmation. Mattern's classic distributed-termination-detection result requires two consecutive rounds of agreement to rule out the ghost-message scenario where a queued event fires between rounds and produces a send that happens to be in flight at exactly the wrong moment.
                                                                                                                                                        
  The existing per-backend implementations of check_for_quiescence were also subtly different — GASNet-EX used asymmetric pre-drain/post-drain sampling, UCX folded a stability check into a single allreduce, MPI used cross-call stability via a static variable — making it hard to reason about correctness uniformly.                                                                                                                                
                  
  Approach                                                                                                                                              
  
  Replace the per-backend check_for_quiescence implementations with a single Mattern's-shaped two-round stability protocol that lives in Network::check_for_quiescence and uses small per-backend sample_quiescence_state / quiescence_allreduce_sum primitives.
                                                                                                                                                        
  The correctness argument is the standard one for Mattern's algorithm:                                                                                 
  
  - A round samples each rank's (any_queue_nonempty, packets_reserved, packets_received, pending_completions) after draining the incoming-message queue.
  - The state is sum-reduced across ranks.
  - Termination is declared when two consecutive rounds observe any_queue_nonempty == 0, pending_completions == 0, and packets_reserved == packets_received globally and the cumulative counters are unchanged from the previous round.                                                          
                                                                                                                                                        
  Stability across two rounds rules out the ghost-message scenario: any in-flight message at the round boundary either changes the counters (failing the equality check between rounds) or leaves a rank with a non-empty queue in the second round (failing the all-quiet check). Correctness depends on the application having promised it will not initiate new work — which is already guaranteed by the shutdown protocol's precondition.                      
                  
  The previous "10-attempt budget" failure rule is replaced with no-progress detection. The runtime keeps iterating as long as the global counters are advancing (slow but progressing systems are not bugs and have no fixed time bound). It aborts only after STUCK_THRESHOLD consecutive iterations during which the counters were frozen and the state was not quiet — which actually identifies a real accounting leak rather than a slow operation.          
                  
  Interface change

  network.h:

  - New Network::QuiescenceStatus enum: DONE (confirmed quiescent), PROGRESSING (counters changing, keep iterating), STUCK (counters frozen but not quiet — real bug).
  - Network::check_for_quiescence now returns QuiescenceStatus instead of bool.                                                                         
  - New NetworkModule::QuiescenceState struct (four uint64_t fields).                                                                                   
  - New pure-virtual NetworkModule::sample_quiescence_state(QuiescenceState&) and NetworkModule::quiescence_allreduce_sum(...).                         
  - Removed pure-virtual NetworkModule::sample_messages_received_count() and NetworkModule::check_for_quiescence(size_t).                               
                                                                                                                                                        
  runtime_impl.cc:                                                                                                                                      
                                                                                                                                                        
  - The shutdown-loop branch on bool done = Network::check_for_quiescence(...) is replaced with a three-way switch on QuiescenceStatus. DONE breaks out, PROGRESSING continues, STUCK aborts with a diagnostic identifying the failure as a real accounting bug rather than a slow operation.
                                                                                                                                                        
  Per-backend changes

  Each backend now provides only the two new primitives. The Mattern's loop itself lives in network.cc and is shared by all backends.                   
  
  GASNet-EX (gasnetex_internal.cc, gasnetex_module.cc/.h, gasnetex_internal.h):                                                                         
                  
  - sample_quiescence_state reports any_queue_nonempty = 1 if any of the four bgwork-item queues (poller, injector, completer, rgetter) is non-empty; sums packets_reserved across all xpairs; reads total_packets_received.load() post-drain; and reports compmgr.num_completions_pending() as pending_completions. The in_progress / work_active "currently executing" flags are deliberately not included — Mattern's two-round stability subsumes what they were guarding against, and including them re-introduces the timing sensitivity that was the demonstrated bug.
  - quiescence_allreduce_sum factors out the existing gex_coll_ireduce + wait_for_full_poll_cycle pattern from the old per-backend check. Behaves identically to before from the network's perspective.                                                                                                 
  
  UCX (ucp_internal.cc, ucp_module.cc/.h, ucp_internal.h):                                                                                              
                  
  - sample_quiescence_state maps outstanding_reqs > 0 to any_queue_nonempty, uses total_msg_sent/total_msg_received for the counters, and computes pending_completions from the remote-completion-sent-vs-received imbalance. The previous "wait_polling on each poller after a failed check" gating moved into sample_quiescence_state so it fires unconditionally on every sample; without this, the new algorithm could spin without giving the pollers a chance to make progress between rounds.
  - quiescence_allreduce_sum wraps the existing UCC_Allreduce.
                                                                                                                                                        
  MPI (mpi_module.cc/.h):
                                                                                                                                                        
  - sample_quiescence_state calls g_am_manager.ensure_polling_progress() and reports messages_sent/messages_rcvd as the counters. MPI's accounting folds remote completions into the same counters as ordinary messages, so pending_completions is reported as 0 and the imbalance shows up directly in the counter check.                                                                                                                                        
  - quiescence_allreduce_sum wraps MPI_Allreduce.
                                                                                                                                                        
  GASNet1 (gasnet1_module.cc/.h):
                                                                                                                                                        
  - The legacy QuiescenceChecker class uses its own multi-step active-message protocol that doesn't expose per-rank counters. Rather than reimplementing it on top of the new interface, sample_quiescence_state reports an always-quiet state so the new algorithm declares termination after two trivial rounds. This matches the spirit of the legacy check (a single global bool); rigorous quiescence detection on the deprecated gasnet1 backend would require exposing the right per-rank counters from the conduit. The legacy QuiescenceChecker class itself is left in place because it's still updated by the gasnet1 AM handlers; cleaning that up is left as a follow-up.

  What this fixes vs. what it doesn't                                                                                                                   
  
  Closes (verified):                                                                                                                                    
  - GASNet-EX timing-sensitive in_progress flag: hole demonstrated in gdb on a frozen failing process; closed because the new design doesn't include any "currently executing" flags in the busy condition.                                                                                                   
  - Fixed 10-attempt budget on both backends: hole demonstrated in CI logs (network still not quiescent after 10 attempts on both GASNet-EX and UCX); closed by the no-progress termination rule.                                                                                                           
  - Single-round stability vulnerability on both backends: theoretical hole, also matches the failing-iteration symptom pattern; closed by the two-round stability check.                                                                                                                                     
                                                                                                                                                        
  Surfaces but doesn't directly close:
  - UCX's outstanding_reqs == 0 is treated as the active flag, but outstanding_reqs only tracks UCX-level requests — it doesn't directly cover Realm-level deferred work that hasn't been submitted to UCX yet. If such deferred work exists, the Mattern's-shaped redesign exposes the gap (the sample_quiescence_state interface is the natural place to enumerate it). Whether this hole is reachable in practice is a follow-up audit — it wasn't  directly demonstrated, only structurally present.                                                                                                                                                               
                                                                                                                                                        
  Files changed                                                                                                                                         
   
  - src/realm/network.h — interface definitions: QuiescenceStatus, QuiescenceState, new virtuals.                                                       
  - src/realm/network.cc — Mattern's two-round loop in Network::check_for_quiescence; LoopbackNetworkModule implementations.
  - src/realm/runtime_impl.cc — shutdown loop uses three-way status switch with no-progress termination.                                                
  - src/realm/gasnetex/gasnetex_module.h/.cc, gasnetex_internal.h/.cc — GASNet-EX implementations.                                                      
  - src/realm/ucx/ucp_module.h/.cc, ucp_internal.h/.cc — UCX implementations.                                                                           
  - src/realm/mpi/mpi_module.h/.cc — MPI implementations.                                                                                               
  - src/realm/gasnet1/gasnet1_module.h/.cc — gasnet1 stub implementations.


elliottslaughter:
The Fuzzer is happy with this as of commit 2038a2915fcb44bcaf0947940e0373e1a4aad94a. I see no failures (at all) in release mode, and only a known failure in debug mode (https://github.com/StanfordLegion/legion/issues/1935). (Edit: I have now completed two full fuzzer runs, with the same results as above.)

I would like someone on the Realm team to go through this with a fine-toothed comb and confirm that it is actually correct. In particular, it is still not obvious to me:

 1. Are all the counters monotonic? We sort of hacked around it, but we still include all the old counters in the sum.
 2. Are the counters being incremented in the right places, with the right conditions? Claude chose some highly unintuitive places or conditions to increment counters and I don't understand the logic well enough to critique the placement in detail.
 3. Is the total counter still necessary? This was added by Claude at my rest for monotonicity, but if we can guarantee that the primary counters are mononotic, we wouldn't need this.

I'm happy with the overall algorithm and it's reassuring that the fuzzer passes on it, but it's a big enough change with enough unintuitive parts to it that I want to be sure before merging.

apryakhin:
I am looking through this PR now

elliottslaughter:
I have also run this through the fuzzer with the UCX backend and it has passed all my tests there too.

Edit to add: my initial tests actually failed to test any network capability, but I belatedly updated my tests to do so, and they have now passed as well.

src/realm/network.cc
```diff
@@ -48,25 +49,162 @@ static void aligned_free(void *ptr)
...
-        size_t messages_received = single_network->sample_messages_received_count();
-        message_manager->drain_incoming_messages(messages_received);
-        return single_network->check_for_quiescence(messages_received);
+
+      // Drain the incoming-message queue first, so that any messages already
+      //  delivered by the network layer have been dispatched to their
+      //  handlers (which may queue more local work, also captured by
+      //  queued_items).  The drain target is messages_to_drain - a strict
+      //  subset of packets_received that excludes wire packets that don't
+      //  pass through IMM (UCX remote-completion replies in particular).
```
apryakhin, line 110: what is IMM here? I am not familiar with this term.

src/realm/network.h
```diff
@@ -180,8 +204,78 @@ namespace Realm {
...
+    //   - packets_reserved exceeding packets_received globally (anything
+    //     in flight on the wire), or
+    //   - pending_completions > 0 (any local-completion bookkeeping that
+    //     could fire and produce comp replies)
+    struct QuiescenceState {
+      // total number of items currently queued on this rank that could
+      //  produce a future network operation: work-item queues (injector,
+      //  completer, poller, rgetter), pending in-flight requests, etc.
+      //  This is a SNAPSHOT (rises and falls); for stability detection it
+      //  must be combined with events_added (monotonic) so that a queue
```
apryakhin, line 255: Why `must be combined with` events_added?

apryakhin: never mind the comment below explains this.

src/realm/network.h
```diff
@@ -180,8 +204,78 @@ namespace Realm {
...
+      //  drain in Network::check_for_quiescence delivered to handlers
+      uint64_t packets_received;
+      // count of pending remote completions on this rank waiting for replies
+      //  to arrive; reduced via SUM.  This field is also a snapshot (rises
+      //  and falls), but its only "decrement-without-monotonic-bump" path
+      //  is comp_reply receipt, which is immediately visible as a change
+      //  to pending_completions itself, so the stability check on this
+      //  field alone is sufficient.
+      uint64_t pending_completions;
+      // local-only (NOT reduced) count of messages this rank has received
```
apryakhin, line 297: I feel that this comments are not made to be read by humans but by AI agents. that reads very robotic. Did you read that yourself?

src/realm/lists.h
```diff
@@ -93,6 +93,10 @@ namespace Realm {
 
     size_t erase(T *entry);
 
+    // O(n) walk - we don't maintain a running size counter, so use this only
```
apryakhin, line 96: I don't understand what this comment means except the linear algorithmic complexity specified.

src/realm/lists.inl
```diff
@@ -156,6 +156,18 @@ namespace Realm {
...
   }
 
+  template <typename T, REALM_PMTA_DECL(T, IntrusiveListLink<T>, LINK), typename LT>
+  inline size_t IntrusiveList<T, LINK, LT>::size(void) const
+  {
+    // Caller is responsible for any external synchronization needed - this
+    //  walks the list head-to-tail without taking the lock.  Intended for
+    //  diagnostic / quiescence-counting use; do not put it on a hot path.
+    size_t n = 0;
+    for(T *p = head.next; p; p = REALM_PMTA_DEREF(p, LINK).next)
```
apryakhin: nit: braces please

src/realm/network.cc
```diff
@@ -48,25 +49,162 @@ static void aligned_free(void *ptr)
...
+      //     required to detect activity in cases where queued_items happens
+      //     to be the same value across rounds while contents flowed through
+      //   packets_reserved/received: cumulative counts; sum across ranks
+      //   pending_completions: count per rank; sum > 0 means at least one
+      //     rank is waiting on a remote completion
+      uint64_t local_arr[5] = {local.queued_items, local.events_added,
+                               local.packets_reserved, local.packets_received,
+                               local.pending_completions};
+      uint64_t total_arr[5] = {0, 0, 0, 0, 0};
+      single_network->quiescence_allreduce_sum(local_arr, total_arr, 5);
```
apryakhin, line 139: what guarantees that in some not so distant future the contract between `quiescence_allreduce_sum` and the counters is not broken? another words I think the assumption here is the implicit that this collective does not contaminate the samples but nothing actually prevents anyone from doing an incorrect change later in either of the backends..for in UCX cases if they don't want to depend on UCC for that.

src/realm/ucx/ucp_internal.cc
```diff
@@ -1699,62 +1700,89 @@ namespace Realm {
...
+      uint64_t rcomp_received = total_rcomp_received.load();
+      state.packets_reserved = msg_sent + rcomp_sent;
+      state.packets_received = msg_received + rcomp_received;
+
+      // pending_completions is unused on UCX - everything that needs
+      //  balancing is already in the reserved/received pair above.
+      state.pending_completions = 0;
+
+      // Only AM-style messages (msg_received) flow through
+      //  IncomingMessageManager; rcomp replies are handled directly by the
```
apryakhin, line 1754: AFAIK this incoming message manager sits above this layer in core runtime. This layer is not aware of this abstraction.

src/realm/ucx/ucp_internal.cc
```diff
@@ -1699,62 +1700,89 @@ namespace Realm {
...
+
+      log_ucp.debug() << "quiescence sample:"
+                      << " queued=" << state.queued_items
+                      << " events_added=" << state.events_added
+                      << " reserved=" << state.packets_reserved << " (msg=" << msg_sent
+                      << "+rcomp=" << rcomp_sent << ")"
+                      << " received=" << state.packets_received
+                      << " (msg=" << msg_received << "+rcomp=" << rcomp_received << ")";
+
+      // Give the poller threads a chance to make progress before the caller
```
apryakhin, line 1769: I don't understand what this means? The "sample_quiescence_state(" name is misleading since that always implies just a read while here we are driving the progress?

src/realm/ucx/ucp_internal.cc
```diff
@@ -1699,62 +1700,89 @@ namespace Realm {
...
+      //  balancing is already in the reserved/received pair above.
+      state.pending_completions = 0;
+
+      // Only AM-style messages (msg_received) flow through
+      //  IncomingMessageManager; rcomp replies are handled directly by the
+      //  UCP completion callback and bypass IMM, so they must NOT be
+      //  included in the drain target.  Including them would deadlock
+      //  drain_incoming_messages since total_messages_handled in IMM never
+      //  counts rcomps.
+      state.messages_to_drain = msg_received;
```
apryakhin, line 1759: I think this design is a bit broken. There is layering smell. The transport that ultimately has no knowledge of the core runtime AM layer attempts to reason about it which is wrong. The cleaner design here will be to let Incoming Message Manager to expose counters that you need for example `pending_handler_count` and it should be accessed in the high level core runtime network layer where the Mattern's algorithm lives.

src/realm/network.h
```diff
@@ -80,9 +80,33 @@ namespace Realm {
     // and a few "global" operations that abstract over any/all networks
     void barrier(void);
 
-    // a quiescence check across all nodes (i.e. has anybody sent anything
-    //  since the previous quiescence check)
-    bool check_for_quiescence(IncomingMessageManager *message_manager);
+    // result of a single quiescence-check round.  The runtime's shutdown loop
+    //  calls Network::check_for_quiescence repeatedly and stops when it sees
+    //  DONE.  We deliberately do not have a "stuck" status: without
```
eddy16112, line 85: The PR description still has the "STUCK" status, given that we do not have this status in the code, it is better to remove it from the description.

src/realm/ucx/ucp_internal.cc
```diff
@@ -1699,62 +1700,89 @@ namespace Realm {
...
+                                  count, UCC_DT_UINT64, UCC_OP_SUM);
       if(rc != UCC_OK) {
-        log_ucp.error() << "allreduce failed in check_for_quiescence";
-        return false;
+        log_ucp.error() << "allreduce failed in quiescence_allreduce_sum";
+        // can't fail gracefully here; return zeros so the caller will see
+        //  imbalance and continue iterating (eventually triggering the STUCK
+        //  detection in Network::check_for_quiescence)
+        for(size_t i = 0; i < count; i++)
+          total_counts[i] = 0;
```
eddy16112: total_counts = 0 could trigger false negative by the caller. The comments shows triggering the STUCK detection, but we do not have the STUCK status now.

src/realm/network.cc
```diff
@@ -48,25 +49,162 @@ static void aligned_free(void *ptr)
...
+      //  emitted, so warnings repeat at a steady cadence rather than
+      //  spamming.
+      static long long next_warn_time_ns = 0;
+      // emit a warning if the counters have been frozen this long.  60 s is
+      //  long enough to outlast any reasonable network round-trip but short
+      //  enough that a CI hang surfaces as actionable log output.
+      static const long long WARN_INTERVAL_NS = 60LL * 1000 * 1000 * 1000;
+    } // namespace QuiescenceCheck
+
+    void reset_quiescence_state(void)
```
eddy16112, line 85: Looks like this function is not called by `RuntimeImpl::wait_for_shutdown`

src/realm/runtime_impl.cc
```diff
@@ -2884,24 +2884,23 @@ namespace Realm {
...
-        }
-
-        if(tries >= 10) {
-          log_runtime.fatal() << "network still not quiescent after " << tries
-                              << " attempts";
-          abort();
-        }
       }
+      if(Network::my_node_id == 0)
+        log_runtime.info() << "quiescent after " << (tries + 1) << " attempts";
```
eddy16112, line 2904: It is better to set the upper limit for the `attempts`, such that we could test the new quiescence algorithm by creating a mock network module.

src/realm/network.cc
lightsighter: Incoming Message Manager

src/realm/network.h
lightsighter: No, like I said in the Realm meeting, this has been vibe coded to see if it even works.

src/realm/lists.h
lightsighter: It is a warning that if you call this function it will require walking the entire list to establish the size. The size of the list isn't kept in the meta-data of the list like it would be in something like a `std::list` where `std::list::size` has O(1).

src/realm/lists.inl
lightsighter: Fixed.

src/realm/network.cc
lightsighter: I think the invariant that protects against that scenario is that the all-reduce sum is done synchronously and therefore it won't be included in the counters itself. It doesn't necessarily have to be done with UCC or some out of band communication, but you do need it to not be counted in the counters. I agree it is a bit strange, but you can't count the quiescence traffic itself in the counters or it will never settle. (The Legion quiescence algorithm has the same issue.) I think it should just be an expected part of the contract between the Realm core and the implementation of `quiescence_allreduce_sum` done by each network backend that they maintain this invariant.

src/realm/ucx/ucp_internal.cc
lightsighter: Yes, but the "rcomp" messages are special to the UCX implementation and don't go through the IncomingMessageManager and this comment is therefore saying why they shouldn't be included in the counts of outstanding messages because they are UCX module level active messages and not Realm active messages.

src/realm/ucx/ucp_internal.cc
lightsighter: It's not actually driving progress, it's just waiting for progress to be made. The background worker threads are the only ones that drive progress. This just ensures that the background worker threads have had a chance to make some progress so the system can settle.

src/realm/ucx/ucp_internal.cc
lightsighter: Same comment as above: the problem is that the UCX backend has special kinds of active messages that are not Realm active messages and do not go through the Incoming Message Manager. Therefore the UCX module needs to think carefully about how to handle its "internal active messages" that it sends for its own control logic with regards to the quiescence algorithm which is what this comment is reflecting.

src/realm/network.h
lightsighter: I've updated the comments to remove references to the STUCK state that no longer exists.

src/realm/ucx/ucp_internal.cc
lightsighter: Also fixed.

src/realm/network.cc
lightsighter: It is called by `check_for_quiescence` which is called by `wait_for_shutdown`.

src/realm/runtime_impl.cc
lightsighter: As the comment above this says, there no longer is a maximum number of attempts because any limit that we might set could have a very slow message that exceeds it (this is how the old protocol would work). The new protocol opts for the alternative approach which is to keep trying until quiescence is achieved. If it is never achieved we'll just be spinning indefinitely and periodically reporting which counts are not summing up properly at which point we should have a pretty good clue as to what is not allowing us to reach quiescence.

apryakhin (approved):

PR #439, opened 2026-05-14
HIP: Sync GPU reduction, fill, and scatter/gather operations with CUDA module

seemamirch:
## Summary

This PR brings the HIP backend's GPU transfer and reduction implementation in line with the CUDA module. It adds a fast GPU reduction path with multi-dimensional affine/transpose layout kernels, fixes fill size reduction and scatter/gather handling, and reorganizes `GPUreduceXferDes` to match CUDA's structure.

### Fast reduction path (`hip_module.h/cc`, `hip_redop.h`, `redop.h`)

- Add `HipRedOpDesc` struct — the HIP equivalent of `CudaRedOpDesc`. Unlike CUDA, HIP launches kernels via `hipLaunchKernel` using host-side `void*` function pointers dir
ectly, so no per-GPU `CUfunction` caching is needed. The descriptor covers 12 kernel variants: basic stride-based (apply/fold × excl/nonexcl), advanced affine (up-to-3D multidimensional layout), and transpose.
- Add `HipModule::register_reduction(Event&, const HipRedOpDesc*, size_t)` — public API for callers to register HIP reduction kernels, equivalent to CUDA's `register_reduction`. Validates proc ownership before registering.
- Add `ReductionKernelsAdvanced` namespace in `hip_redop.h` with `apply_hip_kernel` and `fold_hip_kernel` GPU kernels for affine layout reductions. Remove the old flat stride-based kernels that were superseded.
- Rename `_adv`/`_tran_adv` function pointer fields in `ReductionOpUntyped` to `_advanced`/`_transpose` for consistency with CUDA naming.

### GPUreduceXferDes reorganization (`hip_internal.cc/h`)

- Replace separate `progress_basic_xd` / `progress_custom_xd_data` with a single unified `progress_xd` loop handling both the fast path (GPU kernel launch) and slow path 
(element-by-element), matching CUDA's structure.
- Rename `progress_custom_xd_data` → `fast_reduction_kernel_mode` with updated signature taking explicit port and span parameters passed from the caller.
- Add `describe_kernel_variant(bool is_advanced)` to select the correct advanced/transpose kernel host pointer based on `redop_info.is_fold` / `is_exclusive`.
- Update `setup_redop_kernel` to take `void *redop_args` and build the `params[]` array internally, matching CUDA's interface.
- Rename kernel members for clarity: `kernel_adv` → `kernel_advanced`, `kernel_tran_adv` → `kernel_transpose`; fix `dst_is_ipc` → `src_is_ipc`.

### Fill size reduction fix (`hip_internal.cc`)

- Fix the GPU memset fill size reduction logic: the previous HIP code used an if/else chain with always-false conditions (e.g., `(fill_size >> 1) == 0` instead of `(fill_
size & 1) == 0`), meaning fills were never optimized to smaller memset widths. Ported CUDA's correct loop that iterates from 16 down to 1, finding the largest supported memset size via `(fill_size & (test_fill_size - 1)) == 0`.

### New test (`tests/simple_reduce_gpu.hip`)

- HIP-native GPU reduction registration demonstrating use of `HipRedOpDesc` and `HipModule::register_reduction()` with advanced/transpose kernel variants.
- CMake: add a `REALM_USE_HIP && TEST_USE_GPU` build path for `simple_reduce` that works correctly with shared libs.

## Test plan
- `simple_reduce -all` passes on AMD MI250X (gfx90a)
- HTR reduction test (`reduc.json`, 8 GPUs, 1 node) completes successfully


src/realm/hip/hip_internal.cc
```diff
@@ -48,6 +48,410 @@ namespace Realm {
...
+    {
+      // We don't need a full log2 here
+      for(size_t a = 16; a > 1; a >>= 1) {
+        if((v & (a - 1)) == 0)
+          return a;
+      }
+      return 1; // Unfortunately this can only be byte aligned :(
+    }
+
+    static size_t populate_affine_copy_info(AffineCopyInfo<3> &copy_infos,
```
apryakhin, line 279: @eddy16112 Can we deduplicate all the flow-control here? This looks like an extremely bad decision to just plainly copy-paste that bulky piece of code.

eddy16112: I agreed. It is really hard to maintain the duplicated code. We could try to create an abstraction layer on top of cuda and hip module, and share the common code.

eddy16112: I talked with Sean about this idea when I developed the hip module, but he preferred to keep them separate because CUDA could have new features that are not existed in hip. We can revisit this again now.

muraj: I think we can, but it will take a lot of effort to do I think.  There's a lot of duplicate logic in a lot of places in Realm, especially in the DMA path.  For example, splitting copies such that we lower high dimensional sparse copies into 3D or 1D dense copies is something common to every module but we write the same logic over and over again for each module that supports channels with copies.  This reduction registration and handling is another example.  Re-architecting this will take some time and resources, and we need to develop a plan and allocate resources before we execute.  Please file a github issue to track this and we can discuss further there.

src/realm/hip/hip_internal.cc
```diff
@@ -41,12 +41,419 @@ namespace Realm {
...
+    {
+      assert(mem != nullptr);
+      return pinned_sysmems.count(mem->me) > 0;
+    }
+
+    bool GPU::is_accessible_gpu_mem(const MemoryImpl *mem) const
+    {
+      assert(mem != nullptr);
+      switch(mem->get_kind()) {
+      case Memory::GPU_FB_MEM:
```
eddy16112, line 126: miss the GPU_DYNAMIC_MEM

src/realm/hip/hip_internal.cc
```diff
@@ -1158,204 +2150,506 @@ namespace Realm {
...
-              //   CU_LAUNCH_PARAM_END
-              // };
-              //
-              {
-                AutoGPUContext agc(channel->gpu);
-
-                void *src_ptr = (void *)args->src_base;
-                void *src_device = src_ptr;
+                size_t threads_per_block = 256;
+                size_t blocks_per_grid = 1 + ((elems - 1) / threads_per_block);
```
eddy16112: This is different from cuda? The cuda code is:
```
size_t blocks_per_grid =
                    std::min(1 + ((elems - 1) / threads_per_block),
                             static_cast<size_t>(CUDA_MAX_BLOCKS_PER_GRID));
```

src/realm/hip/hip_module.cc
```diff
@@ -1960,6 +1966,111 @@ namespace Realm {
     */
     }
 #endif
+    static void launch_kernel(const Realm::Hip::GPU::GPUFuncInfo &func_info, void *params,
+                              size_t num_elems, size_t arg_size, GPUStream *stream)
+    {
+      unsigned int num_blocks = 0, num_threads = 0;
+      num_threads = 32;
```
eddy16112: also different from cuda:
```
num_threads = std::min(static_cast<unsigned int>(func_info.occ_num_threads),
                             static_cast<unsigned int>(num_elems));
```

eddy16112 (commented):
I asked the Claude to compare the PR with the CUDA module to check if the changes aligns with the code in CUDA module. Overall the PR LGTM, just minor issues.

src/realm/hip/hip_internal.cc
seemamirch: GPU_DYNAMIC_MEM support in general has not been added to HIP. A bunch of other changes will be needed once it's fully supported

eddy16112: I see. I thought it was already supported.

CMakeLists.txt
```diff
@@ -919,6 +919,72 @@ if(REALM_USE_CUDA)
 endif()
 #endregion
 
+if(REALM_USE_HIP)
+#User - facing arch list
+  set(REALM_HIP_ARCH, ${CMAKE_HIP_ARCHITECTURES})
```
eddy16112: This should be `set(REALM_HIP_ARCH ${CMAKE_HIP_ARCHITECTURES})`, no `,`

src/realm/hip/hip_module.cc
```diff
@@ -43,9 +43,15 @@
...
 #include <dlfcn.h>
+#include <math.h>
+#include <sstream>
+#include <iomanip>
+#include <algorithm>
+#include <utility>
 
 #define IS_DEFAULT_STREAM(stream) ((stream) == 0)
 
+extern unsigned char realm_hip_fatbin[];
```
eddy16112: extern const

src/realm/hip/hip_internal.cc
```diff
@@ -1158,204 +2151,508 @@ namespace Realm {
           out_span_start = out_port->local_bytes_total;
         }
 
-        // have to count in terms of elements, which requires redoing some math
-        //  if in/out sizes do not match
-        size_t max_elems;
-        if(in_elem_size == out_elem_size) {
-          max_elems = max_bytes / in_elem_size;
+#ifdef REALM_USE_HIP_HIJACK
+        bool has_fast_kernels =
```
eddy16112: move it out of the while loop, just as what we did in cuda.

src/realm/hip/hip_internal.cc
```diff
@@ -1100,19 +1993,38 @@ namespace Realm {
...
                               : (redop_info.is_exclusive ? redop->hip_apply_excl_fn
                                                          : redop->hip_apply_nonexcl_fn));
 #ifdef REALM_USE_HIP_HIJACK
       kernel = host_proxy;
+      kernel_advanced = describe_kernel_variant(/*is_advanced=*/true);
+      kernel_transpose = describe_kernel_variant(/*is_advanced=*/false);
 #else
       kernel_host_proxy = host_proxy;
+      kernel_host_proxy_advanced = describe_kernel_variant(/*is_advanced=*/true);
+      kernel_host_proxy_transpose = describe_kernel_variant(/*is_advanced=*/false);
```
eddy16112: It is better to set non-used function pointers to nullptr.

src/realm/hip/hip_internal.cc
```diff
@@ -1100,19 +1993,38 @@ namespace Realm {
...
       GPU *gpu = checked_cast<GPUreduceChannel *>(channel)->gpu;
 
-      // select reduction kernel now - translate to CUfunction if possible
+      // select reduction kernel now - translate to hipfunction if possible
       void *host_proxy =
           (redop_info.is_fold ? (redop_info.is_exclusive ? redop->hip_fold_excl_fn
                                                          : redop->hip_fold_nonexcl_fn)
                               : (redop_info.is_exclusive ? redop->hip_apply_excl_fn
                                                          : redop->hip_apply_nonexcl_fn));
 #ifdef REALM_USE_HIP_HIJACK
```
eddy16112: do we still have hip hijack now?

seemamirch: removed it

eddy16112 (approved):

lightsighter:
@seemamirch Can you update this with the latest `main` branch so we can get it merged?

PR #442, opened 2026-05-15
Fix Instance ID Reuse

lightsighter:
LocalManagedMemory has four "bad-path" branches where a destroy/redistrict is acknowledged from the perspective of release_allocator and future_allocator but cannot yet be applied to current_allocator (because attempt_release_reordering can't satisfy the oldest pending alloc against the new state without breaking fragmentation guarantees). In those branches the inst's PendingRelease was queued with is_ready=true while its tag remained live in current_allocator — and then inst->notify_deallocation() was called unconditionally at the end of the function. That ack runs metadata.initiate_cleanup() → recycle_instance() → release_instance(), which pushes the inst's slot back onto local_instances.free_list.

Once the slot is on the free list, the next MemoryImpl::new_instance() can pop it and hand back the same inst_impl with the same RegionInstance ID. The new caller's allocation request goes into the allocator with that ID — but the allocator still has the old mapping for that same tag from the un-drained PendingRelease. From that point onward, current_allocator.allocated has only one entry per ID (the new one silently overwrites the old via allocated[tag] = idx; in BasicRangeAllocator::allocate, leaking the old range), and the stale PendingRelease will eventually drain and deallocate whichever range is now mapped to the ID — freeing storage that the new instance is actively using.

 The user's two failure signatures are both downstream of this state:

  - BasicRangeAllocator::split_range — assert(allocated.find(new_tags[i]) == allocated.end()) at mem_impl.inl:202. A redistrict whose new instance IDs happen to include a recycled slot finds the slot's old tag still in allocated.
  - BasicRangeAllocator::deallocate — assert(missing_ok) at mem_impl.inl:614. A stale PendingRelease drains after the new instance has been allocated and then properly destroyed, so the tag is no longer in allocated when the old release tries to free it.

The race is narrow (the reporter observed roughly 1 in 70 000 allocations) because it requires triggered + pending_allocs non-empty + attempt_release_reordering can't satisfy the front of pending_allocs, and the recycled slot to be picked up before the original PendingRelease is drained.
  
The Realm semantics that the client calls destroy() exactly once per instance — even for failed allocations — are unchanged. The bug is entirely internal: the destroy ack is correctly fired exactly once, but the slot-recycle effect of that ack happens too early.
  
  The fix

Decouple the dealloc notification from the allocator-state cleanup. A new bool deferred_dealloc_notify field on PendingRelease marks an entry whose tag is still live in current_allocator; the notify_deallocation() call (and the slot recycle it triggers) is deferred until the entry actually drains.
  
  - PendingRelease::deferred_dealloc_notify — set in the four bad-path branches:
  - release_storage_deferrable (triggered && !pending_allocs.empty() && reorder fails)
  - release_storage_immediate (non-oldest entry with !pending_allocs.empty())
  - reuse_storage_deferrable (redistrict, same predicate as the destroy case)
  - reuse_storage_immediate (redistrict, non-oldest with !pending_allocs.empty())

  In each of these sites a local defer_dealloc_notify flag is also set so the function-end inst->notify_deallocation() is
  skipped.
  - attempt_release_reordering gains a std::vector<RegionInstanceImpl*>& out-param. Both of its ready-entry erase loops
 (full-success swap and partial-success swap) append inst pointers of erased entries that carry the flag. The two oldest-path catch-up loops in release_storage_immediate and reuse_storage_immediate do the same as it->release(current_allocator) finally clears the tag.
  - After each function releases allocator_mutex, it walks the collected vector and calls notify_deallocation() on each inst.
  Each inst gets exactly one notification, and only after its tag is actually gone from current_allocator — so the slot is
  never available to new_instance() while the allocator still tracks it.

Regression test

DeferredAllocBadPathTest.ReorderFailDefersInstanceRecycle constructs the minimal scenario that exercises the bad path deterministically:

  1. Memory size 100. Allocate A=30, B=50, C=20 to fill the memory.
  2. release_storage_deferrable(B, e) with e not-yet-triggered — B lands in pending_releases with is_ready=false.
  3. allocate_storage_deferrable(D=50, NO_EVENT) — D fits in the future state (B released) but not the current state, so it
  returns ALLOC_DEFERRED and joins pending_allocs.
  4. release_storage_deferrable(A, NO_EVENT) — triggered destroy, pending_allocs non-empty, release_allocator after A's
  removal only has [0,30) free which can't fit D=50, so attempt_release_reordering returns false: this is the bad path.

  The test then asserts the three invariants the fix preserves:
  - A's PendingRelease is at the back of the queue with is_ready=true && deferred_dealloc_notify=true.
  - A's inst slot is not on local_instances.free_list.
  - A subsequent new_instance() call does not return A's slot.
  
  Step 5 fires B's deferred destroy (by calling release_storage_immediate(B, false) directly — exactly what the DeferredDestroy callback does when the precondition triggers). The oldest-path catch-up drains B, finishes D's allocation, and then drains A as the next consecutive ready entry. The test confirms both queues are empty, D's offset is 30 (where B used to live), and A's slot is now on the free list — proving the deferred notify fired exactly once after A's tag was actually removed from current_allocator.

  Verified both directions: passes with the fix in place (538 unit tests still green), fails on three distinct assertions when the deferred-notify flag is stripped from the targeted bad-path branch.


tests/unit_tests/deferred_alloc_test.cc
```diff
@@ -0,0 +1,208 @@
...
+//   - A's inst slot is NOT yet on local_instances.free_list.
+//   - A fresh new_instance() does not hand A's slot back out.
+//
+// Step 4: fire B's precondition (we bypass the event/DeferredDestroy plumbing
+// by calling release_storage_immediate directly - that's what the deferred-
+// destroy callback does when the precondition triggers).  The oldest-path
+// catch-up drains B, finishes D, and then drains A.  Once A's tag is finally
+// gone from current_allocator the deferred-notify fires exactly once and the
+// slot becomes available for reuse.
+TEST_F(DeferredAllocBadPathTest, ReorderFailDefersInstanceRecycle)
```
eddy16112, line 131: If I understand it correctly, the test only covers the case of release_storage_deferrable+destroy, it is better to add another three cases:
1. release_storage_deferrable + redistrict
2. release_storage_immediate + destroy
3. release_storage_immediate + redistrict

eddy16112 (commented):
The fix LGTM, just add one minor comment for the test.

tests/unit_tests/deferred_alloc_test.cc
lightsighter: I added more tests for those cases. See if they look good to you.

eddy16112 (approved):

PR #443, opened 2026-05-21
DO NOT MERGE: Debugging GASNetEX Module

lightsighter:
Patch summary — six small additions to gasnetex_internal.h/.cc:

  1. gasnetex_internal.h: new FreeSnapshot struct as a member of OutbufMetadata.
  2. gasnetex_internal.cc: #include <sys/syscall.h> and <unistd.h> for gettid.
  3. OutbufMetadata::pktbuf_close: record close_caller PC + counter snapshot at the top.
  4. OutbufManager::free_outbuf: record free_caller PC + thread id at the top.
  5. OutbufManager::alloc_outbuf: record alloc_caller PC on each of the three return paths.
  6. Two close-site instrumentation points (inside the relevant xpair-mutex blocks, recording xpair-side state):
    - reserve_pbuf_helper: close_path=1 + xpair/first_pbuf/cur_pbuf/has_ready snapshot
    - push_packets:advance: close_path=2 + xpair/first_pbuf/cur_pbuf/has_ready snapshot

  On the next fuzzer crash, the gdb command is:

  p cur_pbuf->last_free     # for line 279 crashes (the freed cur_pbuf)
  p new_head->last_free     # for line 2515 crashes (the freed successor)

  That prints all the diagnostic fields. Then info symbol <addr> translates each PC. The combination of:

  - close_path (1 / 2 / 0=neither)
  - close_caller + free_caller PCs (which function/line)
  - counters at close time (was sent==total true? was use_count nonzero?)
  - xpair-side state (was the buffer really first_pbuf? cur_pbuf?)
  - alloc_caller (where was the last alloc from — was it after the free?)

  ...should pin the bug.

### DIFF
--- a/src/realm/gasnetex/gasnetex_internal.cc
+++ b/src/realm/gasnetex/gasnetex_internal.cc
@@ -295,15 +295,22 @@ namespace Realm {
     return true;
   }
 
+  // debug instrumentation: global cross-buffer event ring definitions
+  GlobalEventRecord g_event_ring[GLOBAL_EVENT_RING_SIZE];
+  atomic<uint32_t> g_event_ring_seq{0};
+
   void OutbufMetadata::record_event(EventType type, void *caller_pc, void *xpair,
                                     void *first_pbuf_at, void *cur_pbuf_at,
                                     bool has_ready_at)
   {
+    uint64_t ts = __builtin_ia32_rdtsc();
+    uint64_t tid = static_cast<uint64_t>(syscall(SYS_gettid));
+
     uint32_t idx = event_ring_seq.fetch_add(1) & (EVENT_RING_SIZE - 1);
     EventRecord &ev = event_ring[idx];
-    ev.timestamp = __builtin_ia32_rdtsc();
+    ev.timestamp = ts;
     ev.caller = caller_pc;
-    ev.thread_id = static_cast<uint64_t>(syscall(SYS_gettid));
+    ev.thread_id = tid;
     ev.xpair = xpair;
     ev.first_pbuf = first_pbuf_at;
     ev.cur_pbuf = cur_pbuf_at;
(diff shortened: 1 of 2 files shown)


lightsighter:
Added another patch for debugging:

gasnetex_internal.h — OutbufMetadata now has:
  - An EventType enum (ALLOC / CLOSE / FREE / POP_RESERVE / POP_PUSH).
  - An EventRecord struct (PC, thread id, counter snapshot, xpair-side state).
  - A 16-entry event_ring[] + atomic event_ring_seq for round-robin slot selection.
  - A record_event() member helper.

  gasnetex_internal.cc — five recording points:
  1. pktbuf_close top: EVENT_CLOSE (before assert).
  2. free_outbuf top: EVENT_FREE (before set_state(IDLE)).
  3. alloc_outbuf all three return paths: EVENT_ALLOC (after set_state(PKTBUF)).
  4. reserve_pbuf_helper pop: EVENT_POP_RESERVE with xpair/first_pbuf/cur_pbuf/has_ready snapshot.
  5. push_packets advance: EVENT_POP_PUSH with xpair/first_pbuf-after-store/cur_pbuf/has_ready.
  
  The ring is lock-free (the event_ring_seq.fetch_add(1) & 15 indexing ensures each event goes to a unique slot). Cost per event ≈ one cache-line store
   + one gettid() syscall (~50ns). At 1-in-100K reproduction rate this overhead is invisible.

  On the next repro, the gdb command is:
  
  p cur_pbuf->event_ring
  p cur_pbuf->event_ring_seq

  event_ring_seq.value._M_i & 15 gives the next-write index — so the most recent event is at index (seq - 1) & 15 and they're chronological backwards from there for up to 16 events. Each entry tells us the event type, PC (which we can info symbol to get the source line), thread id, counters at that moment, and xpair state.
  
  The full alloc/close/free/pop history of R will be visible in one dump.

PR #444, opened 2026-05-22
GASNetEX: fix pktbuf_close race that can free an actively-used buffer

lightsighter:
OutbufMetadata::pktbuf_close() reads the non-atomic pktbuf_use_count field twice — once as the argument to remain_count.fetch_add(...), and again in the comparison (prev + pktbuf_use_count) == 0 that decides whether to fire free_outbuf(). Under a specific preemption pattern, the buffer can be legitimately freed and re-allocated between those two reads, causing the comparison to spuriously succeed against a stale prev and a fresh-cycle pktbuf_use_count = 0. The result is a second free_outbuf() call on a buffer that is actively in use by the new cycle, corrupting its state to IDLE mid-flight.

  This manifests as one of two assertion failures, both on a ~1-in-100,000-runs cadence:

  - state == STATE_PKTBUF in pktbuf_reserve / pktbuf_get_offset (a concurrent reservation/commit hits the buffer right after its state was wrongly set to IDLE).
  - new_head->state == STATE_PKTBUF in push_packets (the pusher sees a chained buffer in the wrong state on the next iteration).

  The race

```
  Thread A (pktbuf_close on buffer B, cycle N):
    ...
    read pktbuf_use_count                 -> 16
    remain_count.fetch_add(16)            -> prev=0, remain becomes 16
    *** preempted ***                     between fetch_add and the comparison
                                           (a ~3-instruction window)
```

```
  Other threads, while A is preempted:
    16 dec_usecount() calls drive remain  16 -> 0
    the 16th fires free_outbuf(B)         -> B->state = IDLE
    alloc_outbuf returns B                -> set_state(STATE_PKTBUF)
                                           resets pktbuf_use_count = 0,
                                           remain_count = 0
    reservations/commits accumulate on B  in cycle N+1
```

```
  Thread A resumes:
    re-read pktbuf_use_count              -> 0  (cycle N+1's value)
    check (prev + use_count) == 0         -> (0 + 0) == 0  ← spuriously true
    free_outbuf(B)                        -> B->state = IDLE again,
                                           mid-flight in cycle N+1
```

  The rarity comes from needing the preempt to land in the few instructions between the LOCK XADD and the re-read, and the buffer to complete a full free/realloc cycle before resumption. Both happening together is extremely rare.

  The fix

  Snapshot pktbuf_use_count into a local variable once at the top of pktbuf_close() and use it for both the atomic op and the comparison, in both the overflow and non-overflow branches:

```
  const int saved_use_count = pktbuf_use_count;
  // ... overflow branch:
  int prev = realbuf.load()->remain_count.fetch_add(saved_use_count);
  if((prev + saved_use_count) == 0)
    manager->free_outbuf(realbuf.load());
  // ... non-overflow branch:
  int prev = remain_count.fetch_add(saved_use_count);
  if((prev + saved_use_count) == 0)
    manager->free_outbuf(this);
```
  
  With the snapshot, the comparison correctly reflects whether this close drove remain_count to zero with the original use_count, regardless of whether the buffer was recycled between the atomic op and the comparison.

  Validation

  Reproduced via a long-running fuzzer (~1-in-100K cadence). Root cause was pinned down with per-buffer and global event-ring instrumentation that captured every pktbuf_close / free_outbuf entry along with pktbuf_use_count, remain_count, and state snapshots; the data showed pktbuf_close_entries short by one relative to recorded EVENT_FREE events on the crashing buffer, definitively identifying a free that did not originate from a fresh pktbuf_close call.

### DIFF
--- a/src/realm/gasnetex/gasnetex_internal.cc
+++ b/src/realm/gasnetex/gasnetex_internal.cc
@@ -294,12 +294,19 @@ namespace Realm {
   void OutbufMetadata::pktbuf_close()
   {
     assert(state == STATE_PKTBUF);
+    // snapshot use_count before the atomic op - the field is non-atomic and is
+    //  reset to 0 by set_state(STATE_PKTBUF) when the buffer is recycled.  if
+    //  this thread is preempted after the fetch_add and the buffer completes
+    //  a free/realloc cycle before we resume, a re-read of pktbuf_use_count
+    //  would see the new cycle's 0 and cause (prev + use_count) to compare
+    //  equal to 0 spuriously, freeing a buffer that's actively in use.
+    const int saved_use_count = pktbuf_use_count;
     if(is_overflow) {
       // update the use count on the realbuf, and then we can just delete
       //  ourselves
       assert(realbuf.load());
-      int prev = realbuf.load()->remain_count.fetch_add(pktbuf_use_count);
-      if((prev + pktbuf_use_count) == 0)
+      int prev = realbuf.load()->remain_count.fetch_add(saved_use_count);
+      if((prev + saved_use_count) == 0)
         manager->free_outbuf(realbuf.load());
 
       log_gex_obmgr.info() << "delete overflow: ovbuf=" << this << " baseptr=" << std::hex
@@ -313,8 +320,8 @@ namespace Realm {
     } else {
       // if the result of adding use_count to remain_count is 0, all uses have
       //  already been completed and we can free ourselves
-      int prev = remain_count.fetch_add(pktbuf_use_count);
-      if((prev + pktbuf_use_count) == 0)
+      int prev = remain_count.fetch_add(saved_use_count);
+      if((prev + saved_use_count) == 0)
         manager->free_outbuf(this);
     }
   }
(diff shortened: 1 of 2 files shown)


elliottslaughter:
This passed a very extensive fuzzer run (https://github.com/StanfordLegion/legion/issues/1941#issuecomment-4525914271). I'm happy with it.

Let's proceed to review and merge when ready.

eddy16112:
It LGTM. I have seen a similar pattern in `void OutbufMetadata::databuf_close()`, the `databuf_use_count` will also be set to 0 in `set_state`, but somehow, I did not see the state had been set to the `STATE_DATABUF`, maybe this is a dead code.

lightsighter:
> It LGTM. I have seen a similar pattern in `void OutbufMetadata::databuf_close()`, the `databuf_use_count` will also be set to 0 in `set_state`, but somehow, I did not see the state had been set to the `STATE_DATABUF`, maybe this is a dead code.

That is a good catch. I pushed a fix for that one as well. @eddy16112 can you do the rest of the review for this?

lightsighter:
Actually, you were right the first time, it is dead code so I just removed it.

apryakhin (approved):

eddy16112 (approved):

elliottslaughter (approved):
This has passed a robust set of fuzzer runs, so I'm happy with it.

Issue #438, opened 2026-05-14
Advanced and transpose redop kernels are now required


manopapad:
After https://github.com/StanfordLegion/realm/pull/330, the `CudaModule::register_reduction()` API records the "advanced" & "transpose" optimized [redop kernels](https://github.com/StanfordLegion/realm/blob/main/src/realm/cuda/cuda_redop.h#L237), but doesn't check if these are non-null. Later these are used w/o checks, resulting in the following error:

```
 python: /opt/legate/legate/arch-legate/cmake_build/_deps/realm-src/src/realm/cuda/cuda_internal.cc:2511: bool Realm::Cuda::GPUreduceXferDes::resolve_kernel_slot(Realm::Cuda::GPU*, void*, CUfunc_st*&, CUfunc_st* Realm::Cuda::GPU::GPUReductionOpEntry::*): Assertion `redop->cudaLaunchKernel_fn != 0' failed.
```

Per @lightsighter:

> I at least am not convinced that this should be necessary. I feel like we need to have a discussion about this in the Legion and/or Realm meeting. I at least was under the impression that these new kernels were optional and that if you didn't provide them then Realm would fall back to the older copy paths. However, @muraj tells me that the Realm DMA system doesn't actually do that at the moment and it might be "hard" to do that. I don't think we should expect clients to have to provide these optimized kernels in order for Realm to run correctly so I think we need to discuss which expectations we have of users. Obviously if we make these kernels optional then Legate can still provide them, but I at least don't think they should be required for functional operation.

At the very least there should be a safety check that all required kernels have been provided at registration time.

lightsighter:
I will add this to the discussion for the Legion meeting next week.

> At the very least there should be a safety check that all required kernels have been provided at registration time.

If it is going to be required, then I agree we should enforce the invariant at registration time.

artempriakhin:
> I at least was under the impression that these new kernels were optional and that if you didn't provide them then Realm would fall back to the older copy paths

Yes, I'd agree with that. The least we should do is to enforce the invariant. However, I think we need to discuss if those "advanced" redops should be part of contract and required at all times or made optional. I am leaning towards the "optional" path. This should probably have been implemented from the start since otherwise it changes the application contract and it makes it more "rigid"

elliottslaughter:
Fixed by #445.

PR #445, opened 2026-05-27
Make advanced/transpose CUDA reduction kernels optional in register_reduction

seemamirch:
## Summary

  - Guard `resolve_kernel_slot` against a null `host_proxy` so callers
    can omit the advanced and transpose `CUfunction` slots in
    `CudaRedOpDesc` without crashing.
  - Skip `fast_reduction_kernel_mode` in `progress_xd` when advanced/
    transpose kernels are absent (`has_fast_kernels` guard), falling back
    to the basic stride-based dispatch path instead.
  - Add `cuda_reduce_basic` integration test that registers a same-type
    redop (`LHS == RHS == uint32_t`) with only the 4 basic kernels,
    verifying correct apply and fold results across 1-D, 2-D, and 3-D
    GPU FB reductions.
  - Relates to issue #438

  ## Motivation

  `register_reduction` previously required all 12 kernel variants
  (4 basic + 4 advanced affine + 4 transpose) to be provided.  Callers
  who only have the basic kernels available (e.g. loading a pre-compiled
  cubin via the driver API) had no safe path — passing null advanced/
  transpose slots would crash at dispatch time when
  `fast_reduction_kernel_mode` was entered for equal-size reductions.

  With this change, the advanced and transpose slots in `CudaRedOpDesc`
  are genuinely optional.  When absent, the runtime silently falls back
  to the basic 1-D/2-D stride-based kernel path, which handles all
  dimensionalities correctly via the address cursor.

  ## Test plan

  - [ ] Build with `REALM_ENABLE_CUDA=ON REALM_BUILD_TESTS=ON`
  - [ ] Run `ctest -R cuda_reduce_basic` on a GPU node and confirm pass
  - [ ] Run `ctest -R simple_reduce` to confirm existing full-kernel path is unaffected

### DIFF
--- a/src/realm/cuda/cuda_internal.cc
+++ b/src/realm/cuda/cuda_internal.cc
@@ -653,7 +653,7 @@ namespace Realm {
 
       // 1) Outer loop - iterate over all the addresses for each request
       while(true) {
-        XferPort *in_port = 0, *out_port = 0;
+        XferPort *in_port = nullptr, *out_port = nullptr;
         size_t in_span_start = 0, out_span_start = 0;
         GPU *in_gpu = 0, *out_gpu = 0;
         const GPU::CudaIpcMapping *out_mapping = 0;
@@ -1068,7 +1068,7 @@ namespace Realm {
         const InstanceLayoutPieceBase *in_nonaffine, *out_nonaffine;
 
         size_t in_span_start = 0, out_span_start = 0;
-        XferPort *in_port = 0, *out_port = 0;
+        XferPort *in_port = nullptr, *out_port = nullptr;
         GPU *in_gpu = 0, *out_gpu = 0;
         bool out_is_ipc = false;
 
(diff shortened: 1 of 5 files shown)


src/realm/cuda/cuda_internal.cc
```diff
@@ -2708,8 +2710,11 @@ namespace Realm {
         }
         // add optimized kernels if the condition is satisfied
         // TODO: support different elem sizes
+        bool has_fast_kernels =
+            (kernel_advanced != 0 || kernel_host_proxy_advanced != nullptr) &&
+            (kernel_transpose != 0 || kernel_host_proxy_transpose != nullptr);
```
eddy16112: we could set the value of has_fast_kernels before, such that we do not have to compute it every time in the progress_xd.

muraj (approved):

eddy16112 (approved):

PR #447, opened 2026-06-03
cuda: Fix up base addrs on all gpu channels to translate host address…

muraj:
…es to gpu addresses


eddy16112:
We would have the same bug for replicated heap as well. I think we need to create an issue to keep track of it.

muraj:
> We would have the same bug for replicated heap as well. I think we need to create an issue to keep track of it.

Sure.  Do we have a test that verifies replicated heap?  Because I don't it failing.  The only test I see failing after this change is multiaffine.

PR #451, opened 2026-06-30
Fix DMA Forward Progress

lightsighter:
This fixes a DMA analysis stall where transfer analysis could repeatedly rescan a large sparse index-space domain. TransferDesc::perform_analysis() used domain->volume() during initialization and then called it again before per-field analysis. If the first expensive scan consumed the analyzer’s time quantum, the operation could be requeued before making field progress, then re-enter analysis and repeat the same full sparse-map scan.

The fix caches the exact transfer-domain size once in TransferDesc and reuses it for path sizing, profiling, and IB sizing decisions. This prevents requeued analysis from restarting the expensive IndexSpace::volume() walk.

This also caches indirect-copy domain sizes in IndirectionInfoTyped, avoiding repeated exact-volume scans during gather/scatter path generation.

Changes
  - Added TransferDesc::analysis_domain_size and reuse it after initialization.
  - Replaced repeated domain->volume() calls in transfer analysis with the cached value.
  - Added a lazy cached domain size for IndirectionInfoTyped.
  - Reused a local ind_domain_size in gather/scatter path generation.

Testing: Confirms it fixes a real hang that was occurring when running the Legion CI with the Barnes-Hut application.

### DIFF
--- a/src/realm/transfer/transfer.h
+++ b/src/realm/transfer/transfer.h
@@ -504,6 +504,7 @@ namespace Realm {
     void *fill_data;
     size_t fill_size;
     bool analysis_init_done = false;
+    size_t analysis_domain_size = 0;
     size_t analysis_field_idx = 0;
     size_t analysis_fld_start = 0;
     size_t analysis_fill_ofs = 0;
@@ -657,6 +658,8 @@ namespace Realm {
     IndexSpace<N, T> domain;
     std::vector<IndexSpace<N2, T2>> spaces;
     Channel *addr_split_channel;
+    mutable bool cached_domain_size_valid{false};
+    mutable size_t cached_domain_size{0};
   };
 
   // a TransferOperation is an application-requested copy/fill/reduce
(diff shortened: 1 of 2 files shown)


src/realm/transfer/transfer.h
```diff
@@ -657,6 +658,8 @@ namespace Realm {
     IndexSpace<N, T> domain;
     std::vector<IndexSpace<N2, T2>> spaces;
     Channel *addr_split_channel;
+    mutable bool cached_domain_size_valid{false};
+    mutable size_t cached_domain_size{0};
```
eddy16112: nit: we could use std::optional, then no need for `cached_domain_size_valid`

lightsighter: Fixed.

elliottslaughter:
Legion CI and the fuzzer are happy. Is this ready to go?

eddy16112 (approved):

PR #454, opened 2026-07-02
Fix IB Planning to Avoid Overflows

lightsighter:
Cap and redistribute transfer IB allocations so plans always fit their target memories

Problem

The DMA transfer planner could produce intermediate-buffer (IB) allocation plans that no IB memory could satisfy, crashing at allocation time:

```
ib_alloc: impossible: op=1/0x14079124c260 mem=1a00010000000002 needed=536870904 avail=134217728
```

The planner sized IB edges against the global registered-IB config (core.ib_regmem / -ll:ib_rsize), which is not the actual capacity of the IB memory selected for a given path. For GPU-involved paths this let a planned edge near the 512 MB registered-memory cap target a 128 MB GPU framebuffer IB pool (-ll:ib_fsize) — the exact failure above.

Two deeper issues lurked behind the reported single-edge case:
- Aggregate overflow. A single copy can enumerate many (instance, field) pairs. Fields that don't share both a source and destination instance aren't grouped, so they take independent paths — and those paths can route through the same intermediate memory. The allocator holds all of an op's same-memory IBs simultaneously (allocated as one atomic batch), so N individually-fitting edges can collectively blow the pool.
- Small-transfer over-constraint. A naive per-edge minimum could reject small (sub-256-byte, or single-element) multi-hop transfers that were previously fine.

Fix

Per-edge cap. Each IB edge is sized against min(core.ib_regmem, size_of(that edge's target memory)), so no single edge can exceed the pool it lands in. Edges are sized after their target memory is assigned, so each uses the correct per-memory cap.

Aggregate redistribution. After the full transfer graph is built (all fields/paths), a new pass groups IB edges by memory and, for any memory whose 256-byte-rounded aggregate exceeds capacity, shrinks its edges proportionally to how much each requested above its floor, down to a per-edge minimum. It only aborts when even the minimums can't coexist (genuinely impossible). Shrinking is safe because an IB is a streaming pipeline window, not a whole-transfer buffer — a smaller window just means more, smaller chunks.

Correct progress floor. The per-edge minimum is the raw 2-element ring size (what the engine needs to avoid stalling on the ib_size >> 1 step), capped at the transfer's total data so small domains aren't over-floored, and not rounded up to 256 (the logical ring size needn't be allocator-aligned). This applies to direct copies, reductions, fills, and indirect gather/scatter/scatter-gather edges alike.

Allocator-consistent accounting. IBMemory::do_alloc rounds every request up to 256 bytes; the planner's checks and redistribution now use the same rounding, so a plan the planner accepts is always one the allocator can satisfy (closing a boundary case that could otherwise hang rather than abort).

Overflow-safe sizing. All IB size arithmetic on the planner guard path uses saturating multiply/add and a checked LCM (saturating up to SIZE_MAX, which correctly disables the data-size cap rather than under-shrinking the floor). Replaces an unguarded std::lcm (UB on overflow) and unchecked size * element products.

Single memory-size lookup. Memory capacities are cached in one Memory → size map shared by edge sizing, redistribution, and verification, so each IB memory is queried from the runtime once per planning pass instead of per edge.

Files

- transfer/transfer.cc — per-edge cap, redistribute_ib_sizes_impl, alignment-consistent verify_ib_alloc_sizes, saturating size helpers, memory-size cache
- transfer/ib_memory.{h,cc} — shared IB_ALLOC_ALIGNMENT constant + ib_align_up helper (de-duplicates the bare 256 literals)
- transfer/transfer.h — IBInfo gains min_size and size_granularity fields
- tests/unit_tests/ib_redistribute_test.cc (new) + tests/CMakeLists.txt

Testing

- New unit suite (ib_redistribute_test.cc) covering: fitting plans left untouched, proportional shrink, single-oversized-edge shrink, per-edge granularity preserved, small sub-256-byte transfers preserved, floor capped at data size, overflow saturation, indirect floors capped at natural size, and the impossible-plan / overflow abort paths (death tests).
- Full suites pass: realm_unit_tests (555) and realm_c_unit_tests (148), no regressions.

Notes for reviewers

- No new user-facing flags — IB sizing remains automatic.
- The end-to-end multi-hop/indirect paths that exercise redistribution run under the GPU/multi-node integration tests (CPU-only same-node copies don't stage through an IB); the unit tests lock in the sizing/redistribution math directly.
- clang-format should be run before merge (formatting was deferred during review).

### DIFF
--- a/src/realm/transfer/transfer.cc
+++ b/src/realm/transfer/transfer.cc
@@ -26,6 +26,10 @@
 #include "realm/transfer/ib_memory.h"
 #include "realm/inst_layout.h"
 
+#include <limits>
+#include <map>
+#include <numeric>
+
 #ifdef REALM_ON_WINDOWS
 #include <basetsd.h>
 typedef SSIZE_T ssize_t;
@@ -2997,11 +3001,22 @@ namespace Realm {
     , addrsplit_channel(_addrsplit_channel)
   {}
 
+  // non-static (exposed for unit testing); default lives on this declaration
+  size_t compute_ib_min_size(size_t combined_field_size, CustomSerdezID serdez_id,
+                             size_t max_needed = std::numeric_limits<size_t>::max());
+  size_t compute_ib_natural_size(size_t combined_field_size, size_t domain_size,
+                                 CustomSerdezID serdez_id);
+  static size_t compute_ib_granularity(size_t combined_field_size,
+                                       CustomSerdezID serdez_id);
+  static void init_ib_edge(TransferGraph::IBInfo &ibe, Memory memory, size_t size,
+                           size_t min_size, size_t size_granularity);
+
   static TransferGraph::XDTemplate::IO
   add_copy_path(std::vector<TransferGraph::XDTemplate> &xd_nodes,
                 std::vector<TransferGraph::IBInfo> &ib_edges,
                 TransferGraph::XDTemplate::IO start_edge, const MemPathInfo &info,
-                size_t ib_size = Config::ib_size_bytes)
+                size_t ib_size = Config::ib_size_bytes,
+                size_t ib_min_size = IB_ALLOC_ALIGNMENT, size_t ib_size_granularity = 1)
   {
     size_t hops = info.xd_channels.size();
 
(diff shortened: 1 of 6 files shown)


src/realm/transfer/transfer.cc
```diff
@@ -3982,6 +4156,238 @@ namespace Realm {
...
+                 compute_ib_min_size(combined_field_size, serdez_id, natural_size),
+                 compute_ib_granularity(combined_field_size, serdez_id));
+  }
+
+  static size_t effective_ib_min_size(const TransferGraph::IBInfo &edge)
+  {
+    return (edge.min_size != 0) ? edge.min_size : IB_ALLOC_ALIGNMENT;
+  }
+
+  void redistribute_ib_sizes_impl(std::vector<TransferGraph::IBInfo> &ib_edges,
```
apryakhin, line 4212: I am half way through the review. The overall idea is fine but checking the math takes time.

apryakhin (approved):

PR #462, opened 2026-07-24
Add support for Kokkos v5; drop support for Kokkos v3 (second attempt)

rbberger:
Adds changes on top of #418 to avoid breaking host compilations when host compiler isn't Kokkos compiler (e.g. GCC host compiler, LLVM Kokkos/CUDA compiler).


src/realm/inst_layout.inl
```diff
@@ -1525,13 +1525,21 @@ namespace Realm {
     return v;
   }
 
+#if defined(REALM_USE_KOKKOS_ARRAY_LEGACY)
```
mpokorny, line 1528: I think you can get rid of explicitly naming Kokkos::Array types entirely, as done in https://github.com/StanfordLegion/realm/pull/418/changes#diff-087f7e2304e44361b3cbd843968d38ad56f4d6fb6f8c2ab917201de148345b99L1534
(hmm, the right hand column there). Such a change would reduce the size of this PR in several places.

mpokorny (commented):
As far as this PR only drops the v3 support and adds v5 support, I think that it largely is in accord with #418, with the exception noted below.

src/realm/inst_layout.inl
rbberger: @mpokorny I'll test your new branch. I obviously like your solution more if we can avoid the entire CMake dance.

rbberger:
#418 has resolved my issue in a better way.

PR #463, opened 2026-07-28
Use Cmake HIP Language

lightsighter:
Modernize Realm’s HIP/ROCm build by replacing the custom hipcc --genco invocation with CMake’s native HIP language support.

  - Build Realm’s internal HIP kernels through a HIP object-library target.
  - Let CMake manage HIP compiler selection, architecture flags, language standard, include paths, and source dependencies.
  - Use CMAKE_HIP_ARCHITECTURES as the standard architecture control, removing REALM_HIP_ARCH and manually generated
    --offload-arch flags.

  - Preserve the existing CUDA-like runtime model: convert the device-only output with bin2c and load the embedded module
    through hipModuleLoadData.

  - Generate device-only output with --offload-device-only for AMD Clang and --fatbin for NVIDIA HIP.
  - Remove the explicit hipcc lookup and NVIDIA compiler override.
  - Give CUDA and HIP fatbin generation distinct target names so both backends can be enabled together.
  - Apply Realm’s NDEBUG cleanup to HIP compilation flags.
  - Document HIP platform and architecture configuration in the README.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -464,13 +464,12 @@ if(REALM_ENABLE_HIP)
   set(HIP_PLATFORM ${CMAKE_HIP_PLATFORM})
   find_package(hip REQUIRED CONFIG)
   if(CMAKE_HIP_PLATFORM STREQUAL "nvidia")
-    # nvidia HIP_PLATFORM requies 3.28+ for enable_language(HIP) to work properly
+    # The NVIDIA HIP platform requires CMake 3.28+ for enable_language(HIP).
     cmake_minimum_required(VERSION 3.28 FATAL_ERROR)
     set_target_properties(
       hip::host PROPERTIES INTERFACE_COMPILE_DEFINITIONS "__HIP_PLATFORM_NVIDIA__"
                            INTERFACE_LINK_LIBRARIES CUDA::cudart
     )
-    set(CMAKE_HIP_COMPILER ${CUDAToolkit_NVCC_EXECUTABLE})
   endif()
   if(NOT HIP_ROOT_DIR)
     set(HIP_ROOT_DIR "${ROCM_PATH}")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -464,13 +464,12 @@ if(REALM_ENABLE_HIP)
   set(HIP_PLATFORM ${CMAKE_HIP_PLATFORM})
   find_package(hip REQUIRED CONFIG)
   if(CMAKE_HIP_PLATFORM STREQUAL "nvidia")
-    # nvidia HIP_PLATFORM requies 3.28+ for enable_language(HIP) to work properly
+    # The NVIDIA HIP platform requires CMake 3.28+ for enable_language(HIP).
```
elliottslaughter: I think this is obvious, we should just remove this line.

README.md
```diff
@@ -76,6 +76,21 @@ make -j$(nproc)
...
+      -DREALM_ENABLE_CUDA=OFF \
+      -DREALM_ENABLE_HIP=ON \
+      -DCMAKE_HIP_PLATFORM=amd \
+      -DCMAKE_HIP_ARCHITECTURES=gfx90a
+```
+
+If `CMAKE_HIP_ARCHITECTURES` is omitted, CMake selects an architecture from the
+compiler and available GPU. HIP-on-NVIDIA builds use
+`-DCMAKE_HIP_PLATFORM=nvidia` and require CMake 3.28 or newer.
+
```
elliottslaughter: I'd be inclined to remove this. We're switching to standard CMake functionality, so this is just how CMake works for all HIP projects now. Also, we don't have an equivalent CUDA section, so it's not consistent.

If we want to make a separate GPU documentation section, that's fine. But I wouldn't put it here either.

elliottslaughter (commented):
Looks ok except for a new too-obvious comments that I think we should remove.

We should test this in Frontier CI though.

CMakeLists.txt
lightsighter: Removed

README.md
lightsighter: Removed

lightsighter:
> Looks ok except for a new too-obvious comments that I think we should remove.
> 
> We should test this in Frontier CI though.

I asked both @syamajala and @rbberger to test this out.

elliottslaughter:
I can confirm that our Legion Frontier CI based on this PR is able to build, and does no worse than `master` on tests.

I'd need to spend time cleaning up the Legion `master` branch to fully run this, but I don't think you need to block on me to merge this. However, since I can't run the full test suite through at the moment, probably we should wait until other users confirm to merge this PR.

rbberger:
@elliottslaughter I can not confirm this works for us as long as the Legion part is untouched. See https://github.com/StanfordLegion/legion/issues/1960#issuecomment-5118409831

elliottslaughter:
There is now an accompanying Legion change you can test in https://gitlab.com/StanfordLegion/legion/-/commits/realm-2026-08-07-pr463

elliottslaughter:
Frontier CI is passing in full with the combined branch https://gitlab.com/StanfordLegion/legion/-/commits/realm-2026-08-07-pr463, and @syamajala has also confirmed it works for S3D.

@rbberger Are you able to test on your end? Otherwise I'm inclined to move this along.

rbberger:
@elliottslaughter LGTM. I was able to build and run our FleCSI unit tests with this.

elliottslaughter (approved):
Everyone has tested, so approving now.

PR #464, opened 2026-07-28
Move-Only Logger Message Semantics

lightsighter:
LoggerMessage previously allowed copying, but its copy constructor created an empty buffer while leaving both objects
  active. This could silently lose the copied message content and cause the original object to emit unexpectedly.

  This change makes LoggerMessage a single-owner, move-only type:

  - Deletes copy construction and copy assignment.
  - Adds a move constructor that transfers:
      - The accumulated message content.
      - Stream formatting state.
      - Logger and logging level.
      - Responsibility for emitting the message.

  - Deactivates the moved-from object so it cannot emit during destruction.
  - Deletes move assignment because replacing an existing active message has ambiguous emission semantics.

  This preserves ordinary return-by-value usage under C++17 while turning accidental copies into compile-time errors.
  Existing Legion logger usage remains compatible.

### DIFF
--- a/src/realm/logging.h
+++ b/src/realm/logging.h
@@ -142,7 +142,12 @@ namespace Realm {
     LoggerMessage(Logger *_logger, bool _active, Logger::LoggingLevel _level);
 
   public:
-    LoggerMessage(const LoggerMessage &to_copy);
+    LoggerMessage(const LoggerMessage &to_copy) = delete;
+    LoggerMessage &operator=(const LoggerMessage &to_copy) = delete;
+
+    LoggerMessage(LoggerMessage &&to_move);
+    LoggerMessage &operator=(LoggerMessage &&to_move) = delete;
+
     ~LoggerMessage(void);
 
     template <typename T>
--- a/src/realm/logging.inl
+++ b/src/realm/logging.inl
@@ -367,13 +367,21 @@ namespace Realm {
       stream.construct(buffer.construct());
   }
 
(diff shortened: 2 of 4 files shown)


elliottslaughter:
Can you ask Claude or Codex to microbenchmark this? The last time we made a logger change it broke the compile-time level check, and that's a ~20% performance regression if you get it wrong due to the number of spew messages we compile out.

lightsighter:
> Can you ask Claude or Codex to microbenchmark this? The last time we made a logger change it broke the compile-time level check, and that's a ~20% performance regression if you get it wrong due to the number of spew messages we compile out.

I don't think that's necessary here. There's only 15 lines of code changes here and they are trivially straight-forward to look at and analyze by hand. In fact, this code doesn't actually talk to the logger directly at all. All we've done is change the semantics of the `LoggerMessage` class so that instead of having a copy constructor it has a move constructor. In the case that a move constructor is invoked, it moves the data from the previous `LoggerMessage` to the new `LoggerMessage` without even talking to the Logger directly, so regardless of the logging level, that only matters in the destructor (which is unchanged) when the message is handed off to the logger.

Aside: even though this is a move constructor, we still have to invoke a copy of the message buffer because it could still be in the <160 byte limit that is stored locally in the `LoggerMessage` struct, so we have to copy it. Technically if it is bigger and already stored in the heap we could move it, but seems like the complexity for that is not worth it.

eddy16112 (approved):

PR #465, opened 2026-08-02
Scalable Active Message Multicast

lightsighter:
Motivation

ActiveMessage(NodeSet) was multicast in name only. Each network backend expanded the target set at the source and issued one send per target:

for(NodeID tgt : targets) { ... internal->commit_message(...); }

So a message to N peers cost the originating node N sends, and the target list travelled the wire as a raw node set. This adds a real multicast facility — bounded-radix forwarding over ordinary unicast, with an adaptive target encoding — and deletes the per-target source loops from all four backends.

What changed

A standalone target-set codec — src/realm/multicast.{h,cc}

MulticastTargetSet stores sorted, disjoint, non-adjacent runs, so a range covering thousands of nodes costs one entry and partition() slices it without ever expanding to individual IDs. EncodedMulticastTargets picks among eight wire forms — EMPTY, SINGLE, SMALL_INLINE, RANGES, DELTA_LIST, BITMAP, ALL_NODES, ALL_EXCEPT — by computing the actual serialized byte size of every candidate and taking the minimum, not by a density heuristic. Decoding validates cardinality against both the remaining payload and the configured node count before allocating or looping, plus node bounds, range overflow, canonical ordering and overlong varints; malformed input returns a status rather than aborting, so it stays unit-testable.

This layer deliberately has no dependency on activemsg.h, which is what lets its tests run with no network, no runtime and no message table.

Forwarding, folded into the active-message layer — activemsg.{h,inl,cc}

The origin removes itself from the target set, partitions the remainder into at most R near-equal slices, and sends one envelope to the first node of each. A relay validates it is in its own slice, removes itself, then enqueues child envelopes before invoking the local handler — forward-before-deliver, which matters for messages like runtime shutdown whose handler stops progress. Final delivery presents origin_node as the handler's sender, not the last relay.

Depth is O(log_R M) with O(M) total edges. Handler redispatch reuses the existing ActiveMessageHandlerTable / IncomingMessageManager machinery rather than duplicating signature detection, and large envelopes ride the existing fragmentation path on every hop.

Optional aggregate remote completion uses a transient acknowledgement tree: each relay holds its parent and outstanding-child count, acks once, and immediately reclaims. Fire-and-forget multicasts carry no ack metadata and create no state, and no reusable target plan survives a completed multicast.

Call sites migrated: runtime shutdown (runtime_impl.cc), metadata invalidation (metadata.cc), instance requests (inst_impl.cc), CUDA and HIP IPC.

Removed: ActiveMessage(NodeSet), Network::create_active_message_impl(NodeSet, …), the NetworkModule NodeSet virtuals, the NodeSet recommended_max_payload overloads, and the backend multicast constructors and source loops — 13 files, +51 / −669. Realm::NodeSet itself is unchanged; only the multicast-by-NodeSet path is gone.

Testing

101 new unit tests. Every encoding is forced with an explicit kind assertion, and a shared helper independently recomputes the minimum across all eight candidates to confirm the encoder actually picked it. Coverage includes malformed and fuzzed payloads (verified against a guard page to catch out-of-bounds reads), NodeID 0 and max, partition balance/disjointness/exact-union across radices 1–64, forward-before-deliver ordering, origin-sender preservation across multiple hops, fragmented payloads, and completion firing exactly once.

### DIFF
--- a/src/realm/activemsg.cc
+++ b/src/realm/activemsg.cc
@@ -19,6 +19,9 @@
 #include "realm/atomics.h"
 
 #include "realm/activemsg.h"
+#include "realm/module_config.h"
+#include "realm/multicast.h"
+#include "realm/runtime_impl.h"
 #include "realm/mutex.h"
 #include "realm/cmdline.h"
 #include "realm/logging.h"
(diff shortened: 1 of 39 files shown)


eddy16112:
I do not think we should force every network module to use tree based broadcast. IMHO, we could provide an abstraction function (`bool support_multicast(...)`) to le each network module to report if it could provide native support for multicast, e.g. hardware multicast support. If it is support, we fallback to use radix tree.

lightsighter:
> I do not think we should force every network module to use tree based broadcast. IMHO, we could provide an abstraction function (`bool support_multicast(...)`) to le each network module to report if it could provide native support for multicast, e.g. hardware multicast support. If it is support, we fallback to use radix tree.

To be clear, at the moment, none of our networking layers have native hardware/driver level support for multi-cast active messages which is why the tree broadcast logic all lives on the "client" side of the `Network` module interface in this pull request. There's no custom logic in any of the different `Network` module backends for doing the multicast broadcast tree. The only changes in the different network module backends have been removing some dead code that is no longer necessary with the change to the multicast interface. In the future, if we ever had true hardware level support for multicast then I agree we would need a network module API call for checking for multicast support and pushing that functionality down into the `Network` module backends that support it.

src/realm/activemsg.cc
```diff
@@ -903,4 +906,903 @@ namespace Realm {
...
+      data1 = reinterpret_cast<IncomingMessageManager::CallbackData>(token);
+    }
+
+    // Reuse of the ordinary incoming path is deliberate: it looks the handler up by
+    //  message ID in the same ActiveMessageHandlerTable, honors an inline handler when
+    //  there is time for one, and otherwise queues the message for a handler thread
+    //  with the same TimeLimit semantics as a message that arrived off the wire.  No
+    //  handler-signature detection is duplicated here (plan section 7.4).
+    bool handled = manager->add_incoming_message(
+        sender, msgid, hdr, hdr_size, PAYLOAD_COPY, payload, payload_size, PAYLOAD_COPY,
```
eddy16112, line 1484: The default size of a MessageBlock is 1MB, what will happen if copy (`PAYLOAD_COPY`) a payload whose size is larger than 1MB?

lightsighter: That is a good catch. It was broken for both unicast and multicast. I've pushed fixes to this branch for both cases and had them tested.

src/realm/activemsg.inl
eddy16112, line 258: missing `ensure_chunk_alloc` in this `add_payload`, it is there in the above `add_payload`

eddy16112, line 240: need `ensure_chunk_alloc`

src/realm/activemsg.cc
```diff
@@ -397,32 +442,70 @@ namespace Realm {
...
         }
 
-        bool ok = it->second->add_chunk(frag_info.chunk_id, payload, payload_size);
-        assert(ok);
+        // A callback on a fragment - in practice a remote completion, which rides
+        //  chunk 0 - cannot fire yet: "handled by the target" is only true once the
+        //  whole message has been reassembled and run.  Hold it until then.
+        if(callback_fnptr != nullptr) {
+          it->second.deferred.push_back(
+              DeferredCallback{callback_fnptr, callback_data1, callback_data2});
```
eddy16112: If the completing fragment has a callback and the reassembled message is handled inline, IMM invokes that callback from deferred and then returns true. UCX then invokes the same callback again in [UCPInternal::am_msg_recv_data_ready()](https://github.com/StanfordLegion/realm/blob/08056054a0/src/realm/ucx/ucp_internal.cc#L1191-L1193), potentially releasing the same resources twice. Same issue for other network modules.

eddy16112 (changes requested):

src/realm/activemsg.inl
lightsighter: Fixed

lightsighter: Fixed

src/realm/activemsg.cc
lightsighter: This is fixed generally in the active message abstraction above the networking layers. See if you think it is sufficient or not.

PR #467, opened 2026-08-10
Scalable Barriers (Again)

lightsighter:
Rewrites Realm's barrier internals around two tree-structured protocols designed to scale to thousands of nodes, replacing the all-ranks-message-the-owner scheme. Both protocols were designed and formally verified in TLA+ before implementation; the specs, model-checking scenarios, and design documents ship in tla/barrier/ as the normative record.

Design

- Arrival (BarrierArrive.tla, ARRIVAL_PROTOCOL.md): the owner learns an arrival plan from observed traffic and pushes it out; nodes aggregate cumulative counts up the plan tree, staying O(radix) per node in steady state. Deviations (over-arrival, arrivals at unplanned nodes, run-ahead into future generations) trigger per-generation eager flush so no arrival is ever silently stranded; plan changes use an invalidate/new-plan pair with local deferral to resolve their race. Entirely feed-forward: no timers, no polling, no background work.
- Notification (BarrierNotify.tla, NOTIFICATION_PROTOCOL.md): the owner keeps a subscriber set (no learned tree — the multicast layer plans fan-out per send), publishes version-gated membership, and sends trigger notifications as gap-checked deltas; a node that can't apply a delta pulls. Includes departure hysteresis (K=8, adaptive) so the set tracks waiting patterns without timers.

Verification

Every protocol rule (10 arrival + 8 notification) is backed by a mutation battery: disable the rule, and a model-checking scenario chosen for exactly that failure catches it (tla/barrier/battery.py, 26 rows, ~20 min). The models found real bugs before the code did — including a double-count safety bug where re-parenting a node mid-generation counted its arrivals twice, and several liveness strands in plan-switch races — and drove the fixes (pinned report edges, dead-plan install guards, retroactive outsider signaling). They also surfaced one pre-existing bug: alter_arrival_count is documented as persistent across generations but was applied to a single generation; this PR fixes it to match the documented contract (note: behavioral change, and the API previously had zero test coverage anywhere).

Implementation notes

- Barrier migration is removed; reduction barriers keep the existing eager path unchanged.
- One mutex per BarrierImpl; every spec action is one critical section (documented per-action in STATE_AND_LOCKING.md); has_triggered stays lock-free.
- Deviation-path instrumentation counters dump at destroy_barrier — a run's coverage is checkable, not assumed.

Testing

- tests/barrier_scale_test: a 9-phase deviation driver (over-arrival, outsiders, run-ahead, per-generation plan churn, alters, disjoint waiters, poison, seeded chaos) with counter-based exercise proofs; green at 1/2/4 ranks over MPI, TSAN-clean, full suite 839/839.
- New -ll:barrier_plan_radix (default 8, unchanged behavior) lets small-rank runs form deep plan trees, making the plan-switch race windows reachable on a workstation; with radix 2, the deferral and stale-edge paths execute and pass at 4–6 ranks.
- Remaining scale work (2→128 node ladder, two race-window cases that need ≥8 ranks) is specified in tla/barrier/SCALE_TEST_PLAN.md.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -94,6 +94,8 @@ add_benchmark(lock_chains "${REALM_BENCH_DIR}/lock_chains/lock_chains.cc")
 add_benchmark(lock_contention "${REALM_BENCH_DIR}/lock_contention/lock_contention.cc")
 set(reducetest_ARGS -batches 1) # Run short version
 add_benchmark(reducetest "${REALM_BENCH_DIR}/reducetest/reducetest.cc")
+set(barrier_bench_ARGS -sgens 32 -pgens 128 -reps 1 -adopt 8) # Run short version
+add_benchmark(barrier_bench "${REALM_BENCH_DIR}/barrier_bench/barrier_bench.cc")
 
 add_benchmark(
   task_throughput "${REALM_BENCH_DIR}/task_throughput/task_throughput.cc"
--- a/benchmarks/barrier_bench/barrier_bench.cc
+++ b/benchmarks/barrier_bench/barrier_bench.cc
@@ -91,7 +91,8 @@ namespace TestConfig {
   int reps = 3;     // repetitions (fresh barrier each) per mode+pattern
   int adopt = 32;   // adoption-curve generations printed from serial rep 0
   int pattern = -1; // -1 = both, else a single Pattern
-};
+  int dummy_to_prevent_bad_formatting = 0;
+}; // namespace TestConfig
 
 struct WorkerArgs {
   Barrier barrier;
--- a/benchmarks/barrier_bench/barrier_bench.cc
+++ b/benchmarks/barrier_bench/barrier_bench.cc
@@ -91,7 +91,7 @@ namespace TestConfig {
   int reps = 3;     // repetitions (fresh barrier each) per mode+pattern
   int adopt = 32;   // adoption-curve generations printed from serial rep 0
   int pattern = -1; // -1 = both, else a single Pattern
-}; // namespace TestConfig
+};                  // namespace TestConfig
 
 struct WorkerArgs {
   Barrier barrier;
--- a/benchmarks/barrier_bench/barrier_bench.cc
+++ b/benchmarks/barrier_bench/barrier_bench.cc
@@ -91,7 +91,7 @@ namespace TestConfig {
   int reps = 3;     // repetitions (fresh barrier each) per mode+pattern
   int adopt = 32;   // adoption-curve generations printed from serial rep 0
   int pattern = -1; // -1 = both, else a single Pattern
-};                  // namespace TestConfig
+}; // namespace TestConfig
 
 struct WorkerArgs {
   Barrier barrier;
--- a/benchmarks/barrier_bench/barrier_bench.cc
+++ b/benchmarks/barrier_bench/barrier_bench.cc
@@ -1,0 +1,312 @@
+/*
+ * Copyright 2026 Stanford University, NVIDIA Corporation
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Barrier performance benchmark (tla/barrier/SCALE_TEST_PLAN.md section 4, the
+// T4 A/B comparison).  Uses ONLY the public Barrier API so the identical
+// source compiles against both this branch and main - A/B is by build, never
+// by runtime toggle.  Results go to stdout via printf: the compile-time log
+// floor defaults to WARNING, and a benchmark's output must not depend on the
+// logging configuration.
+//
+// Two timing modes per repetition, each on a fresh barrier:
+//
+//   serial - arrive -> wait -> advance, fully synchronous.  The per-generation
+//     time is one complete arrive/trigger/notify/wake round, measured locally
+//     on rank 0, so it needs no cross-rank clock.  The first generations of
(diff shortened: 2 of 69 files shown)


elliottslaughter:
> alter_arrival_count is documented as persistent across generations but was applied to a single generation; this PR fixes it to match the documented contract

I think the documentation is wrong and the behavior as previously written is intentional. Every barrier generation lives as an independent entity and arrival count adjustments don't flow forward.

I don't know that we care: SCR is dead and we don't maintain any active manual CR codes. I just mention it because there may well be applications that rely on the implemented behavior and I personally never found it unintuitive.

elliottslaughter:
> Barrier migration is removed

Again, maybe we don't care, but SCR did rely on this and perhaps manual CR codes would too.

elliottslaughter:
I'd be curious what the strategy is for keeping the TLA in sync with the C++ (or for that matter, verifying that the ever matched). As far as I know it's a manual process and the formal verification is only as good as your ability to line up the two implementations.

eddy16112:
What is the difference between this PR and Artem's old scalable barrier implementation?

lightsighter:
> > alter_arrival_count is documented as persistent across generations but was applied to a single generation; this PR fixes it to match the documented contract
> 
> I think the documentation is wrong and the behavior as previously written is intentional. Every barrier generation lives as an independent entity and arrival count adjustments don't flow forward.
> 
> I don't know that we care: SCR is dead and we don't maintain any active manual CR codes. 

Note, this pull request does not change the documented semantics. The prior implementation was not faithfully implementing the described semantics. All we're doing here is ensure that the runtime matches the documentation, which in my opinion, is the better semantics for how `alter_arrival_count` should behave.

> I just mention it because there may well be applications that rely on the implemented behavior and I personally never found it unintuitive.

If there was anyone (and I doubt it), they should have reported it as a bug for not matching the documentation. ;)

lightsighter:
> I'd be curious what the strategy is for keeping the TLA in sync with the C++ (or for that matter, verifying that the ever matched). As far as I know it's a manual process and the formal verification is only as good as your ability to line up the two implementations.

If you know of one you let me know. ;) The only one I've got is: ask the LLM to check that the implementation matches the TLA spec. In this case all the code has actually been synthesized from the spec itself (not the other way around) so it's guaranteed to match. We can say something like: always change the spec and then ask an LLM to generate code from it, but there's no way to enforce that.

lightsighter:
> What is the difference between this PR and Artem's old scalable barrier implementation?

I would say that the main way it is different is that it is less buggy. ;) Artem's previous version is still disabled because no one ever went back and fixed all the bugs we were observing with it. This version is getting robustly tested at scale and is passing complex tests with lots of state transitions that exercise all the different states in the TLA+ model, so I have a lot more confidence that it is correct and we'll actually be able to deploy it in practice. Looking at the complexity of the TLA+ models (for both arrivals and notifications) I think it is safe to say that Artem's prior implementation didn't handle all the necessary cases.

elliottslaughter:
> In this case all the code has actually been synthesized from the spec itself (not the other way around) so it's guaranteed to match.

I want to be careful with this claim. This is not a "correctness by construction" argument because the LLM is not a deterministic function. (Neither is a human, for that matter.) There is no guarantee, even if the LLM really does start with the TLA+ code and then derive the C++ from that, that the code is ever going to match. We have at most an empirical argument that this is the case (if we run the code and see no hangs), but the assurance only goes one direction: if we *do* see a hang, we know the C++ code doesn't match (or the TLA+ doesn't describe what we think it describes), but if we *don't* see a hang, we'll never know if they really match 100% or if we're just waiting to discover some corner case.

I still think it's better than before, and TLA+ is better than no TLA+, but I'd be more comfortable if we had a human actually read the new code, because who knows what's hiding in there.

lightsighter:
> > In this case all the code has actually been synthesized from the spec itself (not the other way around) so it's guaranteed to match.
> 
> I want to be careful with this claim. This is not a "correctness by construction" argument because the LLM is not a deterministic function. (Neither is a human, for that matter.) There is no guarantee, even if the LLM really does start with the TLA+ code and then derive the C++ from that, that the code is ever going to match. We have at most an empirical argument that this is the case (if we run the code and see no hangs), but the assurance only goes one direction: if we _do_ see a hang, we know the C++ code doesn't match (or the TLA+ doesn't describe what we think it describes), but if we _don't_ see a hang, we'll never know if they really match 100% or if we're just waiting to discover some corner case.

That may be true in principle, but I gave the LLM extremely strict instructions to ensure a 1-to-1 translation and then had it launch multiple validation agents to confirm the invariant, followed by me asking an entirely different LLM to also perform an audit to confirm the invariant. It's as close to true as it is going to be.

> I still think it's better than before, and TLA+ is better than no TLA+, but I'd be more comfortable if we had a human actually read the new code, because who knows what's hiding in there.

I don't think that's reasonable. There's about 5K lines of code changes and the complexity of the state machines are monstrous. Checking something like 7 nodes with 5 barrier generations and 3 different arrival patterns across those generations requires checking more than 100+M states and required a 24 hour TLC run to verify soundness. You're never going to be able to check it just by looking at the C++ code. I think the only way you can verify this is by running the code. I've done aggressive testing up to 128 nodes (512 ranks with 8 processes/node) on Eos and so far no bugs. Ask your favorite LLM to concoct the most aggressive test that you want and I'll even run it for you. No way is anyone going to find any bugs staring at this C++ code though.

lightsighter:
Here is some performance data showing performance of the `main` branch barriers against these barriers at scale. There are two different patterns: one that is an all-to-all and one that is half-to-half (where half the nodes arrive and the other half subscribe). We can see the new barriers are competitive with the old ones at small node counts and significantly better at scale.

barrier_ab_scaling

lightsighter:
Here is some data for arguing that the default radix for the protocols used in internally should be 8 and that the worst case scenarios are much better for the new barriers.

barrier_radix_tails

lightsighter:
Made another important optimization here which ensures that nodes do not forward on their arrivals in the "expected" case until after they've seen their local arrivals and all their expected child arrivals. Model checked comprehensively with multiple generations and arrival patterns up to 4 nodes, and partial up to 5 nodes. Validated correctness on tests at full scale.

barrier_ab_scaling

abbotts:
I've run some tests for the new barrier on Frontier using four ranks per node (one per NIC). All values are run radix-8 and use the max, mean, and min reported by the barrier_bench benchmark:

image

I've also run the barrier_scale_test at all those data points and it passed. 

I have jobs for 4096 and 8192 ranks in the queue right now but they'll probably take a while to run. I didn't run comparision points against the old barrier, but the new one is looking quite good on Frontier.

PR #468, opened 2026-08-12
Dynamically link CUDA when using HIP NVIDIA platform

elliottslaughter:
HIP apparently defaults to statically linking CUDA when using the NVIDIA platform, and that causes bad things to happen when the application (or Legion) also uses HIP.


elliottslaughter:
Here's the integrated CI job with Legion: https://gitlab.com/StanfordLegion/legion/-/pipelines/2755586365

lightsighter (approved):

PR #475, opened 2026-08-31
Fix mixed-prefix drain in reuse_storage_immediate

lightsighter:
The oldest-entry drain loop applied the offsets-flavor PendingRelease::release() to every entry it drained, but the allocated/offsets outputs are only meaningful for old_inst's own redistrict entry. A ready follower swept up by the drain misbehaved:

 - a plain destroy (empty redistrict_tags) trips the assert(!redistrict_tags.empty()) in debug builds; in release builds split_range returns 0 and clobbers 'allocated', mis-notifying old_inst's already-placed children as ALLOC_EVENTUAL_FAILURE while their tags stay live in current_allocator (leaked ranges)
 - a redistrict follower with more children than old_inst overflows the offsets vector inside split_range (out-of-bounds write)

Apply the offsets-flavor release only to old_inst's own entry and drain followers with the void flavor, matching the equivalent drain in release_storage_immediate.

Adds two DeferredAllocBadPathTest units covering both follower shapes; on the unfixed code they reproduce the two assertion failures exactly.

Adjudicated as BUG-7 in tla/allocation/bugs/BUG-7.md (checked in separately with the deferred-allocation verification campaign).

### DIFF
--- a/src/realm/mem_impl.cc
+++ b/src/realm/mem_impl.cc
@@ -1369,7 +1369,18 @@ namespace Realm {
 
           // catch up the current state
           do {
-            allocated = it->release(current_allocator, offsets);
+            if(it->inst == old_inst) {
+              // only old_inst's own entry uses the offsets flavor -
+              //  'allocated'/'offsets' describe old_inst's new instances and
+              //  must not be clobbered by drained followers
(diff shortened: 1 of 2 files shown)


eddy16112 (approved):
LGTM

PR #476, opened 2026-09-06
Fix Remote Task Spawn Leak

lightsighter:
Every `Realm::Task` created on behalf of a `SpawnTaskMessage` leaks, along with
its argument buffer when `arglen > 64`. `Operation::Operation` hands its single
reference to the finish event via `set_trigger_op`, and `GenEventImpl::trigger`
only released that reference in the `my_node_id == owner` branch. A task spawned
by another node always has a finish event owned by the spawner, so completing it
took the non-owner branch and the reference was stranded. It was released only
if a later remote spawn happened to reuse the same event slot at the next
generation, which dropped the previous operation in `set_trigger_op`.

Observed on a 256-rank S3D run: on one rank, 100,002 of 113,152 remote-spawned
Task objects were never freed (88 MB over ~90 s), and 98% of the ones that were
freed died within 50 us of a new spawn arriving on the same event slot.

A second variant also leaks: `set_trigger_op` required `generation + 1 == gen`,
but a remote proxy's `generation` only advances on local triggers and owner
updates, so a lagging proxy silently declined the operation, leaving its
reference with no holder and making it invisible to `Event::cancel_operation`.

Introduced by fc40bf5e49 ("realm: Reduce lock contention in local task spawning
and completion"), which replaced the `OperationTable` waiter (woken in both
trigger branches) with the `set_trigger_op` handoff.

## Fix

- Track the generation the trigger op belongs to (`current_trigger_op_gen`).
- `trigger()` detaches the op under the mutex in both the owner and non-owner
  branches when the generation matches, and releases it once outside the mutex.
- `set_trigger_op` accepts any `gen > generation` (still asserts `generation + 1`
  on the owner) and returns whether it took ownership. Dropping a previous op is
  now guarded by `REALM_USE_OPERATION_TABLE`, since without the table the op
  releases itself and the unconditional drop was a double release in the race
  window where the trigger message has been sent but local state not yet updated.
- `get_trigger_op` matches on the stored generation, so cancellation can find
  remotely spawned tasks on a lagging proxy.
- `Operation` records whether the event took its reference and, if not, releases
  it itself in `trigger_finish_event` (read into a local before arming the merger,
  since arming may delete the operation).

## Tests

New `GenEventTest` cases in `tests/unit_tests/event_test.cc`, using a minimal
`Operation` subclass that reports its destruction and a mock communicator:

- `RemoteTriggerReleasesTriggerOp`: completing an op on a remotely owned event
  destroys it.
- `RemoteTriggerOpWithLaggingGeneration`: an op for generation 3 on a proxy at
  generation 0 is tracked, found by `get_trigger_op`, and released on trigger.
- `RemoteTriggerOpReplacedBeforeLocalUpdate`: the mock owner spawns the next
  generation's op from inside the trigger send, before local state updates; the
  older op is released and the delayed local update leaves the newer one alone.

Debug build: 561 unit tests and 40 integration tests pass.


src/realm/event_impl.cc
```diff
@@ -1830,6 +1842,16 @@ namespace Realm {
...
 
+        // if an operation on this node (e.g. a task the owner spawned here) was
+        //  responsible for triggering this generation, the event holds its reference -
+        //  drop it just like the owner does above.  Match on the generation rather than
+        //  on 'generation+1': our view of 'generation' can lag, and a newer operation
+        //  may already have been installed for a later generation (see set_trigger_op)
+        if((current_trigger_op != nullptr) && (current_trigger_op_gen == gen_triggered)) {
+          trigger_op_to_release = current_trigger_op;
+          current_trigger_op = nullptr;
+        }
```
eddy16112, line 1865: My agent found an issue here. Here is the case:
1. node A spawns a task on node B.
2. call `Processor::add_finish_event_precondition(Event P)` to add a precondition event P. 
3. poison the P while the task is still running. 
4. Then the completion event (Event C )of the task will be trigged immediately without waiting for the completion of the task. 
5. Node B sends the trigger notification to A, clears `current_trigger_op`, and calls `remove_reference()`. However, the task may not be finished.

lightsighter: I think I pushed a fix for this. See if your agent is happy with it.

eddy16112 (approved):
LGTM

